PR #35, opened 2020-07-21
Redis universal

laraoz:
**Requirements**

- [ ] I have added test coverage for new or changed functionality
- [ ] I have followed the repository's [pull request submission guidelines](../blob/master/CONTRIBUTING.md#submitting-pull-requests)
- [ ] I have validated my changes against all supported platform versions

**Related issues**

Provide links to any issues in this repository or elsewhere relating to this pull request.

Provide a clear and concise description of what you expect to happen.

Provide a clear and concise description of any alternative solutions or features you've considered.

**Additional context**

Add any other context about the pull request here.


eli-darkly:
There are a couple of reasons why we can't accept this PR as is, but thanks for submitting it and we're likely to add an integration for `go-redis` that uses some of this code in some form.

The biggest reason we can't accept the PR is that in the upcoming Go SDK 5.0.0 release (now [in beta](https://github.com/launchdarkly/go-server-sdk/releases)), the database integrations are being moved out of this repository into separate projects. 5.0.0 is also when we are dropping support for older Go versions; 4.x still supports them, and `go-redis` does not. And 5.0.0 still supports code that is built without modules, which `go-redis` does not. So it really has to be a separate project either way.

A secondary reason is that the PR submission guidelines, which have been copied unchanged in your PR description, were not followed here. They state that a PR must have a description by the author, must have test coverage, and must be able to work in all supported platform versions. We welcome contributions, but the minimal standards don't include a PR that has nothing but a title and a diff.

In this case though, please don't take the trouble to revise the PR since it can't go into the v4 code anyway, and the v5 code is fairly different in terms of how database integrations work so it wouldn't be usable as-is there either. We will look at releasing something similar as a separate project, and if that happens we'll mention and link to your PR in the release notes.

eli-darkly:
@laraoz Question: was there any particular reason you chose to use v6 of go-redis, rather than v7 which was released early this year?

redisuniversal/redis.go
```diff
@@ -0,0 +1,229 @@
...
+	if err != nil {
+		loggers.Error("could not ping redis server: %v", err)
+		return nil
+	}
+
+	c := make(chan os.Signal)
+	signal.Notify(c, os.Interrupt, os.Kill)
+	go func() {
+		select {
+		case <-c:
```
eli-darkly, line 41: Due to the select, this is a non-blocking read, which I believe means that this goroutine is going to start, find that there's nothing on the channel, and immediately stop. I think the intention was that it should wait until `signal.Notify` pushes a value.

However, I'm a little uncomfortable with this use of `signal.Notify`. It has a global effect on the process, i.e. the normal behavior of these signals is turned off, so it's not the kind of thing a library would normally do. Could you say more about why it's important to close the Redis client before the process exits for any reason? Normally, exiting a process automatically closes all open sockets, at least in Unix-based OSes. And the go-redis documentation, as far as I can see, doesn't say that it is crucial to call Close in this situation; in fact it says most applications will never need to call Close.

redisuniversal/redis.go
```diff
@@ -0,0 +1,229 @@
...
+			_ = client.Close()
+		}
+	}()
+	if options.MaxRetryCount == 0 {
+		options.MaxRetryCount = defaultRetryCount
```
eli-darkly, line 46: FYI, this was a no-op since `MaxRetryCount` was never being read.

redisuniversal/redis.go
```diff
@@ -0,0 +1,229 @@
...
+	if c.clusterMode {
+		return hashTag + key
+	}
+	return key
+}
+
+// Removing the hashtag from the key
+func (c *redisStoreCore) cleanHashTagKey(key string) string {
+	if c.clusterMode {
+		return strings.Replace(key, hashTag, "",-1)
```
eli-darkly, line 83: I think `TrimPrefix` would have been more appropriate since the operation you're trying to reverse here was specifically adding `hashTag` to the _beginning_ of the string.

redisuniversal/redis.go
```diff
@@ -0,0 +1,229 @@
...
+			if jsonErr != nil {
+				return fmt.Errorf("failed to marshal %s key %s: %s", kind, key, jsonErr)
+			}
+
+			result, err := tx.TxPipelined(func(pipe redis.Pipeliner) error {
+				err = pipe.HSet(baseKey, c.hashTagKey(key), data).Err()
+				if err == nil {
+					result, err := pipe.Exec()
+					// if exec returned nothing, it means the watch was triggered and we should retry
+					if err == nil && len(result) > 0 {
```
eli-darkly, line 160: I believe this was meant to be `len(result) == 0`. As the comment says, a zero-length result means that the watch was triggered, so a non-zero length means that it was not and so the update succeeded.

Running the standard SDK database test suite, as all of the existing database integrations do, would have caught this. I was not able to get updates of existing items to work at all without fixing this.

eli-darkly:
@laraoz - I've added several comments and questions that came up in the process of trying to adapt this code into a separate project that uses the newer SDK model.

redisuniversal/redis.go
eli-darkly: Also, `os.Kill` is not a signal that can be caught.

redisuniversal/redis.go
```diff
@@ -0,0 +1,229 @@
...
+func (c *redisStoreCore) featuresKey(kind ldclient.VersionedDataKind) string {
+	return c.hashTagKey(c.config.Prefix + ":" + kind.GetNamespace())
+}
+
+func (c *redisStoreCore) initedKey() string {
+	return c.config.Prefix + ":" + initedKey
+}
+
+// We use a hashtag in order to keep all keys in the same node (and hash slot) so we can perform
+// and use watch ... exec without issues. Only in ClusterMode
```
eli-darkly, line 72: Is that actually applicable to _hash field_ keys? I thought that if all of the feature flags are stored as hash fields within a single key `launchdarkly:features`, then they are all on the same node anyway.

PR #64, opened 2022-01-06
prepare 5.6.1 release

LaunchDarklyReleaseBot:
**[Note for reviewer: you'll see a lot of things in the diff that don't correspond to any of the changelog notes below. Those are related to build changes that don't affect the functionality of the SDK, such as upgrading the linter, or copyediting of comments.]**

### Changed:
- The SDK will now log a message at `Error` level whenever it detects flag data that should not be possible according to LaunchDarkly's validation rules (for instance, a rule that references a nonexistent variation index, or a circular reference in flag prerequisites). Previously, such messages were only logged at `Warn` level and only if `LoggingConfigurationBuilder.LogEvaluationErrors` was enabled; that is still the default behavior for other kinds of errors (such as "flag not found"), but malformed data is considered a more serious error because it should not be possible and may require assistance from LaunchDarkly support.

### Fixed:
- If the SDK detects a circular reference in flag prerequisites (for example, flag A has flag B as a prerequisite and vice versa), it will now refuse to evaluate the affected flags, returning the default value and an error reason of `MALFORMED_FLAG`. Previously, such a condition would cause infinite recursion and a stack overflow. LaunchDarkly does not allow circular references to be created in flag configurations, but under rare circumstances an SDK could incorrectly perceive such a state if it received updates out of order; or, the SDK could be configured to use data from a file that contained such references.


ldcomponents/http_configuration_builder.go
```diff
@@ -104,11 +106,26 @@ func (b *HTTPConfigurationBuilder) HTTPClientFactory(httpClientFactory func() *h
...
 
+// Header specifies a custom HTTP header that should be added to all requests. Repeated calls to Header with
+// the same key will overwrite previous entries.
+//
+// This may be helpful if you are using a gateway or proxy server that requires a specific header in
+// requests.
+//
+// Overwriting the User-Agent or Authorization headers is not recommended, as it can interfere with communication
+// to LaunchDarkly. To set a custom User Agent, see UserAgent.
+func (b *HTTPConfigurationBuilder) Header(key string, value string) *HTTPConfigurationBuilder {
```
cwaldren-ld, line 124: Should this method be mentioned in the notes?

cwaldren-ld (approved):

Issue #103, opened 2024-02-23
Error log emitted when client connected to relay proxy without events


chradcliffe:
**Is this a support request?** No

**Describe the bug**
When using the `ldcomponents.RelayProxyEndpointsWithoutEvents` to set the `Config.ServiceEndpoints` configuration parameter, the SDK emits an error log (from [here](https://github.com/launchdarkly/go-server-sdk/blob/v6/internal/endpoints/configure_endpoints.go#L86)):
```
ERROR: You have set custom ServiceEndpoints without specifying the Events base URI; connections may not work properly
```

**To reproduce**
Create a client that is connected to a relay proxy without events, e.g.:
```
	relayURI := "http://my-relay-hostname:8080"
	config := ld.Config{
	    ServiceEndpoints: ldcomponents.RelayProxyEndpointsWithoutEvents(relayURI),
	}
```

**Expected behavior**
An error log shouldn't be emitted for a regular configuration option.

**Logs**
N/A

**SDK version**
v6.1.0

**Language version, developer tools**
Go 1.22

**OS/platform**
Linux Alpine

**Additional context**
N/A

cwaldren-ld:
Hi @chradcliffe , I agree this is a bug. Thanks for the report, I'll work on getting that fixed.

_Filed internally as 234826._

cwaldren-ld:
Hi @chradcliffe , I'm targeting the fix at the `v7` major version of the SDK.

I wanted to check if you were planning on moving to v7 in the near term, knowing that v6 is unsupported in ~7 months? 

Otherwise, I will backport it to the v6 branch if you are finding this to be a critical issue.

chradcliffe:
We can bump to v7 so I don't think we need a backport.

cwaldren-ld:
Hi @chradcliffe , this has been fixed in [v7.1.1](https://github.com/launchdarkly/go-server-sdk/releases/tag/v7.1.1).

PR #109, opened 2024-03-05
refactor: remove unused baseURI variables related to ServiceEndpoints

cwaldren-ld:
While working on https://github.com/launchdarkly/go-server-sdk/issues/103, I noticed that some of the code related to `ServiceEndpoints` seems to be using a `baseURI` variable that isn't actually set anywhere. 

I've removed those internal `baseURI` variables in the streaming/polling/event builders, which allowed me to remove an unused `overrideValue` parameter from some of the ServiceEndpoint utility functions.

### DIFF
--- a/internal/endpoints/configure_endpoints.go
+++ b/internal/endpoints/configure_endpoints.go
@@ -48,11 +48,8 @@ func getCustom(serviceEndpoints interfaces.ServiceEndpoints, serviceType Service
 }
 
 // IsCustom returns true if the service endpoint has been overridden with a non-default value.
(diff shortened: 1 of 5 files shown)


internal/endpoints/configure_endpoints.go
```diff
@@ -48,11 +48,8 @@ func getCustom(serviceEndpoints interfaces.ServiceEndpoints, serviceType Service
 }
 
 // IsCustom returns true if the service endpoint has been overridden with a non-default value.
-func IsCustom(serviceEndpoints interfaces.ServiceEndpoints, serviceType ServiceType, overrideValue string) bool {
```
cwaldren-ld, line 51: `overrideValue` was always `""`.

keelerm84 (approved):

PR #110, opened 2024-03-05
fix: do not emit error when RelayProxyEndpointsWithoutEvents is used

cwaldren-ld:
When `RelayProxyEndpointsWithoutEvents` is used to configure an SDK's `ServiceEndpoints`, it throws an [error-level log message](https://github.com/launchdarkly/go-server-sdk/issues/103) explaining that only some URIs are set.

This logic was intended to surface a common mistake of only setting one URI when the developer meant to set all 3. In this case the error is incorrect as we're making this choice explicitly as a supported use-case: use Relay for streaming and polling, but send events to LaunchDarkly.

There are many ways to approach fixing this in a backwards compatible manner. Ideally, we'd have constructors that embody the supported use-cases. 

To avoid introducing a new construction API, I've taken a minimal approach of adding a new setter/getter on the `ServiceEndpoint` struct to indicate if partial-specification of custom URIs was intentional or not. 

This way, we can later check that the constructed `ServiceEndpoints` intentionally hasn't set the events endpoint and decide whether or not to log a message.


interfaces/service_endpoints.go
```diff
@@ -7,9 +7,31 @@ package interfaces
 // values such as Streaming, or use the helper method
 // [github.com/launchdarkly/go-server-sdk/v7/ldcomponents.RelayProxyEndpoints].
 //
+// Important note: if one or more URI is set to a custom value, then all URIs should be set to custom values.
+// Otherwise, the SDK will emit an error-level log to surface this potential misconfiguration, while using default values for
+// the unset URIs.
+//
+// There are some scenarios where it is desirable to set only some of the fields, but this is not recommended for general
+// usage. If you're scenario requires it, you can call [WithPartialSpecification] to suppress the
```
cwaldren-ld:
```suggestion
// usage. If your scenario requires it, you can call [WithPartialSpecification] to suppress the
```

interfaces/service_endpoints.go
```diff
@@ -2,14 +2,41 @@ package interfaces
...
+// You may set individual values such as Streaming, or use the helper method
 // [github.com/launchdarkly/go-server-sdk/v7/ldcomponents.RelayProxyEndpoints].
 //
+// Important note: if one or more URI is set to a custom value, then
+// all URIs should be set to custom values. Otherwise, the SDK will emit
+// an error-level log to surface this potential misconfiguration, while
+// using default values for the unset URIs.
+//
+// There are some scenarios where it is desirable to set only some of the
+// fields, but this is not recommended for general usage. If you're scenario
```
keelerm84:
```suggestion
// fields, but this is not recommended for general usage. If your scenario
```

keelerm84 (approved):

PR #117, opened 2024-03-26
ci: fix sed command in check-go-versions.yml

cwaldren-ld:
See created PR: https://github.com/launchdarkly/go-server-sdk/pull/127

### DIFF
--- a/.github/workflows/check-go-versions.yml
+++ b/.github/workflows/check-go-versions.yml
@@ -25,8 +25,8 @@ jobs:
       - name: Parse officially supported Go versions
         id: parse
         run: |
-          echo "latest=${{ fromJSON(env.fetch-api-data)[0].latest }}" >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


.github/workflows/check-go-versions.yml
```diff
@@ -25,8 +25,8 @@ jobs:
       - name: Parse officially supported Go versions
         id: parse
         run: |
-          echo "latest=${{ fromJSON(env.fetch-api-data)[0].latest }}" >> $GITHUB_OUTPUT
-          echo "penultimate=${{ fromJSON(env.fetch-api-data)[1].latest }}" >> $GITHUB_OUTPUT
```
cwaldren-ld, line 29: Instead of using the full Go versions (like 1.22.1), just grab only the "cycle" (1.22). This is what we want in the docs and in the go.mod.

kinyoklion (approved):

PR #125, opened 2024-03-26
feat: Implement supporting types for hooks.

kinyoklion:
This PR should be reviewed before other hooks PRs.

This PR adds supporting types for hooks, but does not use them from the client.


internal/hooks/hook_iterator.go
```diff
@@ -0,0 +1,52 @@
+package hooks
+
+import (
+	"github.com/launchdarkly/go-server-sdk/v7/ldhooks"
+)
+
+type hookIterator struct {
```
kinyoklion: Hook iterator provides a forward/reverse iterator for iterating a collection of hooks.

kinyoklion:
I am going to make a couple changes.
1. Skip execution when no hooks registered.
2. ~~`ldvalue.CopyArbitraryValue(defaultValue)` is allocating. If I run the benchmark with that line I get 1 alloc per op. If I remove it I get 0. So I need to look at that.~~ Removed the need for this.

internal/hooks/hook_iterator.go
```diff
@@ -0,0 +1,52 @@
...
+	reverse    bool
+	cursor     int
+	collection *[]ldhooks.Hook
```
cwaldren-ld: Any reason to store a pointer to the slice? (is it so new stuff can be added at runtime?)

kinyoklion: I can change this to not be a pointer. I did a few experiments and didn't undo this along the way.

internal/hooks/hook_iterator.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+func newHookIterator(reverse bool, hooks *[]ldhooks.Hook) *hookIterator {
+	cursor := -1
```
cwaldren-ld: As a general comment, since the package is called `hooks` it would be implied that iterator/runner type names don't need the  "hook" prefix. 

For the iterator since it's private, probably doesn't matter. But for the runner, perhaps `hooks.Runner`  / `hooks.NewRunner()` would be cleaner.

internal/sharedtest/hooks.go
```diff
@@ -0,0 +1,169 @@
...
+
+// HookStage is the stage of a hook being executed.
+type HookStage string
+
+const (
+	// HookStageBeforeEvaluation is the stage executed before evaluation.
+	HookStageBeforeEvaluation HookStage = "before"
+	// HookStageAfterEvaluation is the stage executed after evaluation.
+	HookStageAfterEvaluation = "after"
+)
```
cwaldren-ld: I always found it ambiguous if the second const is actually a raw string or the alias. (I forget which.)

So I do:
```suggestion
const (
	// HookStageBeforeEvaluation is the stage executed before evaluation.
	HookStageBeforeEvaluation = HookStage("before")
	// HookStageAfterEvaluation is the stage executed after evaluation.
	HookStageAfterEvaluation = HookStage("after")
)
```

ldhooks/hook_metadata.go
```diff
@@ -0,0 +1,33 @@
+package ldhooks
+
+// HookMetadata contains information about a specific hook implementation.
+type HookMetadata struct {
```
cwaldren-ld: Suggestion: `HookMetadata` -> `Metadata`, so you have `ldhooks.Metadata` & `ldhooks.NewMetadata`

internal/sharedtest/hooks.go
```diff
@@ -0,0 +1,169 @@
...
+			for i, afterCall := range localAfterCalls {
+				if reflect.DeepEqual(afterCall, call.EvalCapture) {
+					localAfterCalls = slices.Delete(localAfterCalls, i, i+1)
+					found = true
+				} else {
+					logDebugData(t, afterCall, call)
+				}
+			}
+		default:
+			assert.FailNow(t, fmt.Sprintf("Unhandled hook stage: %v", call.HookStage))
```
cwaldren-ld: Is there not an assertion that takes the format string? Perhaps not

ldhooks/evaluation_series_context.go
```diff
@@ -0,0 +1,46 @@
+package ldhooks
+
+import (
+	"github.com/launchdarkly/go-sdk-common/v3/ldcontext"
+	"github.com/launchdarkly/go-sdk-common/v3/ldvalue"
+)
+
+// EvaluationSeriesContext contains contextual information for the execution of stages in the evaluation series.
+type EvaluationSeriesContext struct {
```
cwaldren-ld, line 9: Whenever one of these value receivers is invoked, it'll need to copy I guess 2 pointers + sizeof(Context) + sizeof(Value).

ldhooks/evaluation_series_data.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+// Get gets the value associated with the given key. If there is no value, then ok will be false.
+func (b EvaluationSeriesData) Get(key string) (value any, ok bool) {
+	val, ok := b.data[key]
+	return val, ok
+}
+
+// AsAnyMap returns a copy of the contents of the series data as a map.
+func (b EvaluationSeriesData) AsAnyMap() map[string]any {
```
cwaldren-ld, line 29: https://pkg.go.dev/maps#Clone

kinyoklion:
> I am going to make a couple changes.
> 
> 1. Skip execution when no hooks registered.
> 2. ~`ldvalue.CopyArbitraryValue(defaultValue)` is allocating. If I run the benchmark with that line I get 1 alloc per op. If I remove it I get 0. So I need to look at that.~ Removed the need for this.

I am not sure there would actually be any tangible benefit to 1, at least with the ability to dynamically register hooks.

In node I was able to avoid the micro-task queue, but in Go I would incur the lock to check regardless and that would be the most expensive part.

ldhooks/evaluation_series_data.go
cwaldren-ld: I think this isn't an immutable map. Since the caller could call `AsAnyMap()` and then modify the values, which are shallow cloned. 

E.g.
```golang
foo := data.AsAnyMap()
foo["bar"].someField = 3

assert.Equal(data["bar"].someField, 3)
```

maybe?

ldhooks/evaluation_series_data.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+}
+
+// Set sets the given key to the given value.
+func (b *EvaluationSeriesDataBuilder) Set(key string, value any) *EvaluationSeriesDataBuilder {
+	b.data[key] = value
+	return b
+}
+
+// SetFromMap copies the keys and values from the given map to the builder.
```
cwaldren-ld: perhaps "merges" might give the right idea - otherwise they may think it starts from empty

ldhooks/hooks.go
```diff
@@ -0,0 +1,93 @@
...
+// contain an implementation of all series and handler interfaces. It should not contain the Hook interface directly
+// because the implementer should be required to implement GetMetadata.
+
+// A Hook is used to extend the functionality of the SDK.
+//
+// In order to avoid implementing unused methods, as well as easing maintenance of compatibility, implementors should
+// compose the `UnimplementedHook`.
+//
+//	type MyHook struct {
+//	  UnimplementedHook
```
cwaldren-ld:
```suggestion
//	  ldhooks.UnimplementedHook
```

ldhooks/hooks.go
```diff
@@ -0,0 +1,93 @@
...
+//	}
+type Hook interface {
+	GetMetadata() HookMetadata
```
cwaldren-ld:
```suggestion
	Metadata() HookMetadata
```

(no Get prefix is idiomatic)

ldhooks/hooks.go
```diff
@@ -0,0 +1,93 @@
...
+// Hook implementors should use this to avoid having to implement empty methods and to ease updates when the Hook
+// interface is extended.
+//
+//	type MyHook struct {
+//	  UnimplementedHook
+//	}
+//
+// The hook should implement at least one stage/handler as well as the GetMetadata function.
+type UnimplementedHook struct {
+}
```
cwaldren-ld:
```suggestion
type Unimplemented struct {}
```

That way we can have `ldhooks.Unimplemented`.

ldhooks/evaluation_series_data.go
kinyoklion: Unless I am misunderstanding. The map itself is immutable. You cannot add new keys to it and you cannot replace values in it.

The values themselves may or may not be mutable depending on their type. Which is how I anticipate our implementations being in any language. (Which is also the purpose, calling `span.end()` for instance on a span we get out of the map.

You cannot change the original input map itself though (add keys to it, assign different values to keys).

cwaldren-ld: No, you're right, I was under the temporary illusion that I had used LdValue in .NET (or you had specified that).

ldhooks/hooks.go
```diff
@@ -0,0 +1,93 @@
...
+// contain an implementation of all series and handler interfaces. It should not contain the Hook interface directly
+// because the implementer should be required to implement Metadata.
+
+// A Hook is used to extend the functionality of the SDK.
+//
+// In order to avoid implementing unused methods, as well as easing maintenance of compatibility, implementors should
+// compose the `Unimplemented`.
+//
+//	type MyHook struct {
+//	  Unimplemented
```
cwaldren-ld:
```suggestion
//	  ldhooks.Unimplemented
```
Mainly because that is exactly what they will need to type so they can copy/paste

ldhooks/hooks.go
```diff
@@ -0,0 +1,93 @@
...
+// hookInterfaces is an interface for implementation by the Unimplemented
+type hookInterfaces interface {
+	EvaluationSeries
+}
+
+// Unimplemented implements all Hook methods with empty functions.
+// Hook implementors should use this to avoid having to implement empty methods and to ease updates when the Hook
+// interface is extended.
+//
+//	type MyHook struct {
```
cwaldren-ld, line 62: ditto

cwaldren-ld (approved):

PR #126, opened 2024-03-26
feat: Add support for hooks.

kinyoklion:
Second PR.

This connects hooks to the ldclient.

It does not add the `Ex` methods to any interface, and it does not add `AddHooks` to any interface.


ldclient_migration_test.go
```diff
@@ -1,61 +1,97 @@
...
-func TestDefaultIsReturnedIfFlagEvaluatesToNonStringType(t *testing.T) {
-	flag := ldbuilders.NewFlagBuilder("migration-key").Build() // flag is off and we haven't defined an off variation
+type MigrationVariationMethod = func(
+	client *LDClient,
+	key string,
+	context ldcontext.Context,
+	stage ldmigration.Stage,
+) (ldmigration.Stage, interfaces.LDMigrationOpTracker, error)
+
+func TestMigrationVariation(t *testing.T) {
```
kinyoklion, line 25: Re-worked these tests to be parameterized so they could be ran against `MigrationVariation` and `MigrationVariationEx`.

ldclient.go
```diff
@@ -24,14 +25,43 @@ import (
...
 	"github.com/launchdarkly/go-server-sdk/v7/ldcomponents"
+	"github.com/launchdarkly/go-server-sdk/v7/ldhooks"
 	"github.com/launchdarkly/go-server-sdk/v7/subsystems"
 	"github.com/launchdarkly/go-server-sdk/v7/subsystems/ldstoreimpl"
 )
 
 // Version is the SDK version.
 const Version = internal.SDKVersion
 
+const boolVarFuncName = "LDClient.BoolVariation"
```
cwaldren-ld: Can surround this with a giant `const` block

ldclient.go
```diff
@@ -678,7 +732,40 @@ func (client *LDClient) BoolVariationDetail(
...
+	return detail.Value.BoolValue(), detail, err
+}
+
+// BoolVariationEx is the same as [LDClient.BoolVariation], but accepts a context.Context.
+//
+// Returns defaultVal if there is an error, if the flag doesn't exist, or the feature is turned off and
+// has no off variation.
+//
+// For more information, see the Reference Guide: https://docs.launchdarkly.com/sdk/features/evaluating#go
+func (client *LDClient) BoolVariationEx(
```
cwaldren-ld: If the `Ex` methods are only differentiated by taking a Context, would you consider `BoolVariationCtx` instead as the convention?

kinyoklion: I considered `Context` and `Ctx` which I think make sense if your original function didn't also take a thing called `Context`. When I started calling them it seemed a little strange.

We could also add functional options right now as well, which would make them different, but I cannot think of any options to actually add. So they would be placeholder.

I could change to Ctx if you prefer.

cwaldren-ld: To my Go-brain, Ctx seems intuitive and I would prefer it. If I was unfamiliar with the LD concept of context I'd immediately know this took a `context.Context`.

cwaldren-ld: Side note: I think it'd be good to explain what the context is used for here. Like, the programmer might think "Ok, this takes a context, it must be a long running operation. Therefore if I want to end evaluation early, I should call context.Cancel()`.

cwaldren-ld (approved):

ldclient.go
kinyoklion: I did make a try at this, then I wondered if it would be a problem. If we decide we want to use the context for something, will it become a problem. 

Maybe it wouldn't be a problem to specifically say that cancelling a context does not cancel an evaluation. Versus saying specifically what we use it for.

PR #131, opened 2024-03-29
chore(refactor): Remove AddHooks, refactor execution, add contract tests.

kinyoklion:
This is a "chore" because I plan on rebase merging this feature branch and this is not a release-note change.


internal/hooks/evaluation_execution.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	gocontext "context"
+
+	"github.com/launchdarkly/go-sdk-common/v3/ldlog"
+	"github.com/launchdarkly/go-sdk-common/v3/ldreason"
+	"github.com/launchdarkly/go-server-sdk/v7/ldhooks"
+)
+
+// EvaluationExecution represents the state of a running series of evaluation stages.
+type EvaluationExecution struct {
```
kinyoklion, line 12: I changed the model so the runner runs an execution. This should be slightly less copying of the parameters as well.

ldclient.go
```diff
@@ -1144,18 +1162,6 @@ func (client *LDClient) WithEventsDisabled(disabled bool) interfaces.LDClientInt
 	return client.withEventsDisabled
 }
 
-// Generic method for evaluating a feature flag for a given evaluation context.
```
kinyoklion, line 1147: If the method doesn't exist, then it cannot accidentally be called.

cwaldren-ld (approved):

PR #148, opened 2024-05-08
chore: Add a warning message if the waitFor duration is too long

louis-launchdarkly:
This adds a warning-level message and we make sure the test still runs

BEGIN_COMMIT_OVERRIDE
fix: add warning log message if client init duration is longer than recommended (60s)
END_COMMIT_OVERRIDE

### DIFF
--- a/ldclient.go
+++ b/ldclient.go
@@ -34,8 +34,8 @@ import (
 // Version is the SDK version.
 const Version = internal.SDKVersion
 
-// HighWaitForSeconds is the initialization wait time threshold that we will log a warning message
-const HighWaitForSeconds = 60
+// highWaitForSeconds is the initialization wait time threshold that we will log a warning message
+const highWaitForSeconds = 60
 
 const (
 	boolVarFuncName   = "LDClient.BoolVariation"
@@ -333,8 +333,8 @@ func MakeCustomClient(sdkKey string, config Config, waitFor time.Duration) (*LDC
 
 		// If you use a long duration and wait for the timeout, then any network delays will cause
 		// your application to wait a long time before continuing execution.
-		if waitFor.Seconds() > HighWaitForSeconds {
-			loggers.Warnf("Make Client was called with a timeout greater than %d. "+
+		if waitFor.Seconds() > highWaitForSeconds {
+			loggers.Warnf("Client was created was with a timeout greater than %d. "+
 				"We recommend a timeout of less than %d seconds", HighWaitForSeconds)
 		}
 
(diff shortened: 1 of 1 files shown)


ldclient.go
```diff
@@ -34,6 +34,9 @@ import (
 // Version is the SDK version.
 const Version = internal.SDKVersion
 
+// HighWaitForSeconds is the initialization wait time threshold that we will log a warning message
+const HighWaitForSeconds = 60
```
cwaldren-ld:
```suggestion
// highWaitForSeconds is the initialization wait time threshold that we will log a warning message
const highWaitForSeconds = 60
```
Uppercase = public variable. Totally intuitive, right?

Anyways, I think we want this private unless there is a good reason to expose it.

ldclient.go
```diff
@@ -327,6 +330,14 @@ func MakeCustomClient(sdkKey string, config Config, waitFor time.Duration) (*LDC
 	if waitFor > 0 && client.dataSource != datasource.NewNullDataSource() {
 		loggers.Infof("Waiting up to %d milliseconds for LaunchDarkly client to start...",
 			waitFor/time.Millisecond)
+
+		// If you use a long duration and wait for the timeout, then any network delays will cause
+		// your application to wait a long time before continuing execution.
+		if waitFor.Seconds() > HighWaitForSeconds {
```
cwaldren-ld:
```suggestion
		if waitFor.Seconds() > highWaitForSeconds {
```

ldclient.go
```diff
@@ -327,6 +330,14 @@ func MakeCustomClient(sdkKey string, config Config, waitFor time.Duration) (*LDC
...
+			loggers.Warnf("Make Client was called with a timeout greater than %d. "+
+				"We recommend a timeout of less than %d seconds", HighWaitForSeconds)
```
cwaldren-ld: I suspect this will only substitute the seconds for %d once, you might have to add the parameter twice

ldclient.go
```diff
@@ -327,6 +330,14 @@ func MakeCustomClient(sdkKey string, config Config, waitFor time.Duration) (*LDC
 	if waitFor > 0 && client.dataSource != datasource.NewNullDataSource() {
 		loggers.Infof("Waiting up to %d milliseconds for LaunchDarkly client to start...",
 			waitFor/time.Millisecond)
+
+		// If you use a long duration and wait for the timeout, then any network delays will cause
+		// your application to wait a long time before continuing execution.
+		if waitFor.Seconds() > HighWaitForSeconds {
+			loggers.Warnf("Make Client was called with a timeout greater than %d. "+
```
cwaldren-ld:
```suggestion
			loggers.Warnf("Client was created was with a timeout greater than %d. "+
```
So the wording works for MakeClient and MakeCustomClient

ldclient.go
louis-launchdarkly: You are right, I tested in the Go playground but was only using one %d before. This is a case where writing a simple string is probably better than the format. I added the argument now.

cwaldren-ld (approved):
Approved assuming output was verified

PR #152, opened 2024-05-28
Additional test for Broadcaster data race

cwaldren-ld:
This is a small cleanup to the existing Broadcasters implementation to use `slices.DeleteFunc`.

Additionally, I've upped the number of parallel goroutines used in the new race detection test to 10, to have a slightly higher chance of catching issues, and added a new test that randomly runs the functions against each-other.

### DIFF
--- a/internal/broadcasters.go
+++ b/internal/broadcasters.go
@@ -1,6 +1,7 @@
 package internal
 
 import (
+	"slices"
 	"sync"
 )
 
@@ -21,7 +22,7 @@ type Broadcaster[V any] struct {
 }
 
 // We need to keep track of both the channel we use for sending (stored as a reflect.Value, because Value
-// has methods for sending and closing), and also the
+// has methods for sending and closing), and also the channel for receiving.
 type channelPair[V any] struct {
 	sendCh    chan<- V
 	receiveCh <-chan V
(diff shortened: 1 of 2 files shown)


internal/broadcasters.go
```diff
@@ -48,27 +49,23 @@ func (b *Broadcaster[V]) AddListener() <-chan V {
...
 }
 
 // HasListeners returns true if there are any current subscribers.
 func (b *Broadcaster[V]) HasListeners() bool {
 	b.lock.RLock()
-	hasListeners := len(b.subscribers) > 0
-	b.lock.RUnlock()
-	return hasListeners
+	defer b.lock.RUnlock()
+	return len(b.subscribers) > 0
```
cwaldren-ld, line 68: Explicitly locking/unlocking is probably a tiny bit faster, but I like the defer pattern here.

internal/broadcasters_test.go
```diff
@@ -97,7 +98,7 @@ func TestBroadcasterDataRace(t *testing.T) {
 		func() { b.HasListeners() },
 		func() { b.RemoveListener(nil) },
 	} {
-		const concurrentRoutinesWithSelf = 2
+		const concurrentRoutinesWithSelf = 10
```
cwaldren-ld: This change is mainly based on intuition and some experience, but I don't have any actual hard data to back up "more goroutines = higher chance of catching the race".

internal/broadcasters.go
JohnStarich: That was my first thought as well, but the contribution guidelines told me not to 😉

internal/broadcasters_test.go
```diff
@@ -83,7 +84,7 @@ func testBroadcasterGenerically[V any](t *testing.T, broadcasterFactory func() *
 	})
 }
 
-func TestBroadcasterDataRace(t *testing.T) {
+func TestBroadcasterDataRaceSelf(t *testing.T) {
```
JohnStarich: This test actually already launches all of the functions in parallel as well. Both in parallel with themselves and with each other. I figure the new test is redundant with this one?

internal/broadcasters_test.go
JohnStarich: I’m uncertain on best practice there. I was able to reliably reproduce both race conditions with 2, but it should be enough to cause a flaky test at worst. 10 seems fine, I’m just not sure if it adds anything 🤷

internal/broadcasters.go
cwaldren-ld: Ah, yea, I see that note. That's specifically referring to hot paths like the evaluation engine, but I see how it's rather ambiguous.

internal/broadcasters_test.go
cwaldren-ld: I'll revert this change.

internal/broadcasters_test.go
cwaldren-ld: You're right. I'm not sure how much value add there is to explicitly running them in random order like the new test, because perhaps the go-routine scheduler already does that.

Seems fine to err on the side of >= 0 more coverage though.

PR #153, opened 2024-05-29
fix: resolve data race in Broadcasters system

cwaldren-ld:
This resolves a data race in the Broadcaster implementation, specifically ` HasListeners()` was not protected by a lock. 

Additionally it swaps out the mutex for a `RWLock` so that concurrent readers don't block each other. 

A behavioral change is that the `Broadcast` method now locks the subscribers array, so that `Add` or `RemoveListener` cannot be called concurrently.

### DIFF
--- a/internal/broadcasters.go
+++ b/internal/broadcasters.go
@@ -2,8 +2,6 @@ package internal
 
 import (
 	"sync"
-
-	"golang.org/x/exp/slices"
 )
 
 // This file defines the publish-subscribe model we use for various status/event types in the SDK.
@@ -19,7 +17,7 @@ const subscriberChannelBufferLength = 10
 // Broadcaster is our generalized implementation of broadcasters.
 type Broadcaster[V any] struct {
 	subscribers []channelPair[V]
-	lock        sync.Mutex
+	lock        sync.RWMutex
 }
 
 // We need to keep track of both the channel we use for sending (stored as a reflect.Value, because Value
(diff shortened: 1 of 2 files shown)


tanderson-ld (approved):

internal/broadcasters.go
```diff
@@ -50,35 +49,31 @@ func (b *Broadcaster[V]) AddListener() <-chan V {
...
 // Broadcast broadcasts a value to all current subscribers.
 func (b *Broadcaster[V]) Broadcast(value V) {
-	b.lock.Lock()
-	ss := slices.Clone(b.subscribers)
-	b.lock.Unlock()
-	if len(ss) > 0 {
-		for _, ch := range ss {
-			ch.sendCh <- value
-		}
+	b.lock.RLock()
```
kinyoklion, line 73: So, there is a behavioral change here, which may or may not be relevant. But previously during the broadcast you could remove a listener (the lock wasn't retained during the actual broadcast), and now you wouldn't be able do get the exclusive lock during that time.

cwaldren-ld: hmm, we probably shouldn't hold the lock during the broadcast as that could take an unbounded amount of time to complete. I'll revert that and add a test.

cwaldren-ld: Ok, so there's a problem here: if we allow removing a listener while Broadcast is running concurrently, then it could try to send to a channel that is closed (due to removing the listener.) This is a panic. 

So that behavior was already bad. 

I'm thinking it's better to say "if you aren't draining your receivers, you're not gonna be able to add or remove receivers" which would be the case if we locked around the broadcast.

cwaldren-ld: I'm reverting to locking around the broadcast. I'll note in the release that this may affect the use-case of adding/removing a listener while broadcasting.

PR #186, opened 2024-08-29
chore: Create FDv2 compatible datasource implementation

keelerm84:
Introduces the new `datasourcev2` package to hold all FDv2 related
functionality.

This package contains a streaming datasource implementation that is
compatible with our current SDK, but pulls data updates from a
FDv2-compatible source.

It does not support any of the other FDv2-specific features like picking
up from a known state. That will be introduced in later work as we
re-shape the datasource integration on a larger scale.

### DIFF
--- a/internal/datasourcev2/helpers.go
+++ b/internal/datasourcev2/helpers.go
@@ -1,0 +1,54 @@
+package datasourcev2
+
+//nolint: godox
+// TODO: This was copied from datasource/helpers.go. We should extract these
+// out into a common module, or if we decide we don't need these later in the
+// v2 implementation, we should clean this up.
+
+import (
+	"fmt"
+
+	"github.com/launchdarkly/go-sdk-common/v3/ldlog"
+)
+
+// Tests whether an HTTP error status represents a condition that might resolve on its own if we retry,
+// or at least should not make us permanently stop sending requests.
+func isHTTPErrorRecoverable(statusCode int) bool {
+	if statusCode >= 400 && statusCode < 500 {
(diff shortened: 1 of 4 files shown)


internal/datasourcev2/helpers.go
```diff
@@ -0,0 +1,54 @@
+package datasourcev2
```
keelerm84, line 1: @cwaldren-ld Had to remove the underscore. The linter did not care for that.

internal/datasourcev2/helpers.go
```diff
@@ -0,0 +1,54 @@
...
+
+//nolint: godox
```
keelerm84, line 3: Since this is an evolving bit of work, I'm just silencing my todo notes. I personally think it's completely reasonable to have that in the code as you develop stuff, but if you don't agree, I can pull it out and track it separately.

internal/datasourcev2/streaming_data_source.go
```diff
@@ -0,0 +1,679 @@
...
+
+	// It would be nice if we had the same value available in both so we could
+	// use that as the key consistently throughout the the process.
+	ID     string `json:"id"`
+	Target int    `json:"target"`
+	Code   string `json:"code"`
+	Reason string `json:"reason"`
+}
+
+// This is the general shape of a put-object event. The delete-object is the same, with the object field being nil.
```
keelerm84, line 91: Left these in for completeness.

cwaldren-ld (approved):

PR #187, opened 2024-08-29
chore: Create FDv2 compatible polling data source

keelerm84:
Expands the `datasourcev2` package with a polling datasource implementation that is compatible with our current SDK, but pulls data updates from a FDv2-compatible source.

It does not support any of the other FDv2-specific features like picking up from a known state. That will be introduced in later work as we re-shape the datasource integration on a larger scale.


internal/datasourcev2/polling_data_source.go
```diff
@@ -0,0 +1,201 @@
+package datasourcev2
```
keelerm84, line 1: As last time, most of this is a straight copy from the older `datasource` package. The only changes are in the requester, so theoretically we could use all the existing stuff and just replace that piece.

However, since we are planning to evolve this work pretty quickly, I figured the duplication is likely preferable.

I will call out where the actual changes occur.

internal/datasourcev2/polling_http_request.go
```diff
@@ -0,0 +1,151 @@
...
+	} else if len(data) != 1 {
+		return nil, false, malformedJSONError{errors.New("missing expected put event")}
+	}
+
+	putData, ok := data[0].(datasource.PutData)
+	if !ok {
+		return nil, false, malformedJSONError{errors.New("payload is not a PutData")}
+	}
+
+	return putData.Data, cached, nil
```
keelerm84, line 106: This handling of the body response is the only bit I've changed so far. This polling implementation isn't doing anything yet with the version, not does it deal with changesets at this point.

internal/datasourcev2/streaming_data_source.go
```diff
@@ -65,58 +65,6 @@ const (
 // Otherwise, the client initialization method may time out but we will still be retrying in the background, and
 // if we succeed then the client can detect that we're initialized now by calling our Initialized method.
 
-type changeSet struct {
```
keelerm84, line 70: Just moved to a new types file.

internal/datasourcev2/types.go
```diff
@@ -0,0 +1,87 @@
...
+
+// Event returns the name of the event.
+func (e event) Event() string {
+	return e.Name
+}
+
+// Data returns the raw data of the event.
+func (e event) Data() string {
+	return string(e.EventData)
+}
```
keelerm84, line 33: Added these types so we could take the polling payload and treat it like a stream of events for processing in the functions I made in the other PR.

cwaldren-ld (approved):

testservice/go.sum
```diff
@@ -15,25 +15,21 @@ github.com/launchdarkly/ccache v1.1.0 h1:voD1M+ZJXR3MREOKtBwgTF9hYHl1jg+vFKS/+VA
 github.com/launchdarkly/ccache v1.1.0/go.mod h1:TlxzrlnzvYeXiLHmesMuvoZetu4Z97cV1SsdqqBJi1Q=
 github.com/launchdarkly/eventsource v1.6.2 h1:5SbcIqzUomn+/zmJDrkb4LYw7ryoKFzH/0TbR0/3Bdg=
```
cwaldren-ld, line 16: was this supposed to be udpated?

PR #188, opened 2024-09-06
refactor: encapsulate existing SDK data system [DRAFT]

cwaldren-ld:
The purpose of this PR is to encapsulate all of the configuration/plumbing logic related to "fdv1" store/sources. 

This will live behind a `DataSystem` interface defined in the client. It should have no behavioral change. 

We can then introduce the fdv2 concepts satisfying this same interface. At that point, the client's configuration will determine which data system to pick at runtime. This will allow both to coexist for a testing phase. 

Current status: This PR is like a sandbox. It's pretty huge and touches a lot of systems. We can factor out some of the components and introduce them incrementally, given we are happy with how they fit together in this PR.


internal/datastore/data_store_status_provider_impl.go
```diff
@@ -2,19 +2,33 @@ package datastore
 
 import (
 	"github.com/launchdarkly/go-server-sdk/v7/interfaces"
-	"github.com/launchdarkly/go-server-sdk/v7/subsystems"
 )
 
+type StatusMonitorable interface {
```
cwaldren-ld, line 7: This refactoring defines an interface with a single method - one of DataStore's methods. This component only needed that single method; this interface documents that fact. 

It also makes it _way easier_ to pass in something to satisfy it, instead of having to implement a bunch of no-op methods.

internal/datasystem/fdv1_datasystem.go
```diff
@@ -0,0 +1,146 @@
+package datasystem
```
cwaldren-ld, line 1: This whole file implements the status quo, pre-FDv2. 

It refactors out a bunch of config logic that used to live in the SDK client, stuffing it behind the new `DataSystem` interface. 

Since the `DataSystem` interface offers some new capabilities, it also implements those capabilities in terms of the old system. I'll call that out in the comments.

internal/datasystem/fdv1_datasystem.go
```diff
@@ -0,0 +1,146 @@
...
+	dataSourceStatusProvider    interfaces.DataSourceStatusProvider
+	dataStoreStatusBroadcaster  *internal.Broadcaster[interfaces.DataStoreStatus]
+	dataStoreStatusProvider     interfaces.DataStoreStatusProvider
+	flagChangeEventBroadcaster  *internal.Broadcaster[interfaces.FlagChangeEvent]
+	dataStore                   subsystems.DataStore
+	dataSource                  subsystems.DataSource
+}
+
+func NewFDv1(dataStoreFactory subsystems.ComponentConfigurer[subsystems.DataStore], dataSourceFactory subsystems.ComponentConfigurer[subsystems.DataSource], clientContext *internal.ClientContextImpl) (*FDv1, error) {
+	system := &FDv1{
```
cwaldren-ld, line 24: This is basically a copy-paste of existing SDK configuration logic. If it looks like a bunch of circular dependencies, that's because it is.

internal/datasystem/fdv1_datasystem.go
```diff
@@ -0,0 +1,146 @@
...
+	if factory == nil {
+		// COVERAGE: can't cause this condition in unit tests because it would try to connect to production LD
+		factory = ldcomponents.StreamingDataSource()
+	}
+	contextCopy := *context
+	contextCopy.BasicClientContext.DataSourceUpdateSink = dataSourceUpdateSink
+	return factory.Build(&contextCopy)
+}
+
+func (f *FDv1) DataSourceStatusBroadcaster() *internal.Broadcaster[interfaces.DataSourceStatus] {
```
cwaldren-ld, line 89: This series of getters is to satisfy the SDK client's public interface, where these status providers are exposed. The FDv2 data system will also need to return something here, but they may not be functional (for now, or maybe I'll find a way to make them function.)

internal/datasystem/fdv1_datasystem.go
```diff
@@ -0,0 +1,146 @@
...
+	if f.flagChangeEventBroadcaster != nil {
+		f.flagChangeEventBroadcaster.Close()
+	}
+	return nil
+}
+
+func (f *FDv1) Offline() bool {
+	return f.dataSource == datasource.NewNullDataSource()
+}
+
```
cwaldren-ld, line 135: Here's a new DataSystem method: the "data status"!


This gives users what they really want to know: does the SDK have data, and if so, is it "fresh".  In the FDv1 world, we can map the statuses like so:

1. If the data source is initialized, then the data has been `Refreshed` (because our data sources are either LaunchDarkly-provided streaming or polling, we know that if they initialized, we have fresh data)
2. If the data source is not initialized, then we may have some `Cached` data from a persistent store.
3. Otherwise we got nothing. 

We used to have a bool (initialized), now we're offering a tribool.

internal/datasystem/fdv1_datasystem.go
```diff
@@ -0,0 +1,146 @@
...
+	if f.dataSource.IsInitialized() {
+		return Refreshed
+	} else if f.dataStore.IsInitialized() {
+		return Cached
+	} else {
+		return Defaults
+	}
+}
+
+func (f *FDv1) Store() subsystems.ReadOnlyStore {
```
cwaldren-ld, line 149: There's no need for the SDK client to have write access to the store, I think. Returning a more restricted interface here to represent that.

internal/datasystem/fdv2_datasystem.go
```diff
@@ -0,0 +1,253 @@
...
+	bcasters := &broadcasters{
+		dataSourceStatus: internal.NewBroadcaster[interfaces.DataSourceStatus](),
+		dataStoreStatus:  internal.NewBroadcaster[interfaces.DataStoreStatus](),
+		flagChangeEvent:  internal.NewBroadcaster[interfaces.FlagChangeEvent](),
+	}
+
+	dataStoreUpdateSink := datastore.NewDataStoreUpdateSinkImpl(bcasters.dataStoreStatus)
+	clientContextCopy := *clientContext
+	clientContextCopy.DataStoreUpdateSink = dataStoreUpdateSink
+	clientContextCopy.DataSourceUpdateSink = store
```
cwaldren-ld: Some funkyness here just like in `FDv1` struct due to the way `clientContext` is passed between components during the configuration stage.

internal/datasystem/fdv2_datasystem.go
```diff
@@ -0,0 +1,253 @@
...
+		primarySync:   cfg.Synchronizers.Primary,
+		secondarySync: cfg.Synchronizers.Secondary,
+		offline:       cfg.Offline,
+		loggers:       clientContext.GetLogging().Loggers,
+		broadcasters:  bcasters,
+	}
+
+	if cfg.Store != nil {
+		fdv2.dataStoreStatusProvider = datastore.NewDataStoreStatusProviderImpl(cfg.Store, dataStoreUpdateSink)
+		store.SetPersistent(cfg.Store, cfg.StoreMode, fdv2.dataStoreStatusProvider)
```
cwaldren-ld: Why not pass `cfg.Store`, `cfg.StoreMode`, and `dataStoreStatusProvider` in the constructor of `store`? We can't.. because the `dataStoreStatusProvider` depends on `cfg.Store` being built, but we can't build _that_ without `store` (because `store` implements `DataSourceStatusSink`)

Sadness.

internal/datasystem/fdv2_datasystem.go
```diff
@@ -0,0 +1,266 @@
...
+}
+
+func (f *FDv2) launchTask(task func()) {
+	f.wg.Add(1)
+	go func() {
+		defer f.wg.Done()
+		task()
+	}()
+}
+
```
cwaldren-ld, line 130: Here's the main workflow. Initialize, then synchronize.

config.go
```diff
@@ -197,4 +197,41 @@ type Config struct {
...
+	//
+	//    config := ld.Config{
+	//        DataSystem: ldcomponents.DataSystem().PollingOnly(),
+	//    }
+	//
+	// If you'd like to load data from a local source to provide redundancy if there is a problem
+	// connecting to LaunchDarkly, you can add a custom initializer:
+	//
+	//    config := ld.Config {
+	// 	  		DataSystem: ldcomponents.DataSystem().PrependInitializers(myCustomInitializer),
```
keelerm84, line 232: Instead of making this prepend, can this be a simple add style method? Prepending adds an unnecessary "think of this as a stack" mental overhead I think we'd like to avoid. 

I realize these are supposed to run before our default initializer, but maybe we can figure out a better way to represent that. Maybe we call these custom initializers, and the docs make that clear that's before the built-in. We can discuss.

Also we should make this method singular unless it is going to be variadic.

internal/datasourcev2/streaming_data_source.go
```diff
@@ -115,17 +119,25 @@ func NewStreamProcessor(
 	return sp
 }
 
+//nolint:revive // DataInitializer method.
+func (sp *StreamProcessor) Name() string {
+	return "StreamingDataSourceV2"
+}
+
+func (sp *StreamProcessor) Fetch(ctx context.Context) (*subsystems.InitialPayload, error) {
```
keelerm84, line 127: Should we not even have this method implemented then?

internal/datasourcev2/streaming_data_source.go
```diff
@@ -161,6 +173,8 @@ func (sp *StreamProcessor) consumeStream(stream *es.Stream, closeWhenReady chan<
 
 			sp.logConnectionResult(true)
 
+			// TODO(cwaldren/mkeeler): Should this actually be true by default? It means if we receive an event
```
keelerm84, line 176: I think leaving it as is would result in the same behavior we currently have, right? So it's probably fine to keep it.

We could consider setting it to false until we get a useful event. But practically I doubt it would make much of a difference.

internal/datasourcev2/streaming_data_source.go
```diff
@@ -183,21 +197,18 @@ func (sp *StreamProcessor) consumeStream(stream *es.Stream, closeWhenReady chan<
...
-					// scenario 2a in error handling comments at top of file
-				} else {
-					sp.loggers.Errorf("Failed to store %s in data store; will restart stream until successful", updateDesc)
-					shouldRestart = true // scenario 2b
-					processedEvent = false
-				}
+				// TODO: the data source previously had the responsibility of figuring out if storing an update failed,
+				// and then potentially restarting the streaming connection to get a new PUT so that it could init
+				// the database if updates got lost. This is no longer the responsibility of the data source (now the
+				// data system will handle it.) Ideally, the update sink's methods should not be fallible.
```
keelerm84, line 210: Vital changes.

```suggestion
				// data system will handle it). Ideally, the update sink's methods should not be fallible.
```

internal/datasystem/fdv2_datasystem.go
```diff
@@ -0,0 +1,302 @@
...
+func (f *FDv2) runPersistentStoreOutageRecovery(ctx context.Context, statuses <-chan interfaces.DataStoreStatus) {
+	for {
+		select {
+		case newStoreStatus := <-statuses:
+			if newStoreStatus.Available {
+				// The Store has just transitioned from unavailable to available
+				if newStoreStatus.NeedsRefresh {
+					f.loggers.Warn("Reinitializing data Store from in-memory cache after after data Store outage")
```
keelerm84, line 165:
```suggestion
					f.loggers.Warn("Reinitializing data Store from in-memory cache after data Store outage")
```

internal/datasystem/fdv2_datasystem.go
```diff
@@ -0,0 +1,302 @@
...
+	f.runSynchronizers(ctx, closeWhenReady, payloadVersion)
+}
+
+func (f *FDv2) runPersistentStoreOutageRecovery(ctx context.Context, statuses <-chan interfaces.DataStoreStatus) {
+	for {
+		select {
+		case newStoreStatus := <-statuses:
+			if newStoreStatus.Available {
+				// The Store has just transitioned from unavailable to available
+				if newStoreStatus.NeedsRefresh {
```
keelerm84, line 164: Cleaner to make compound conditional instead? Accepting this would require another change below, and probably some formatting fixes.

```suggestion
			if newStoreStatus.Available && newStoreStatus.NeedsRefresh {
				// The Store has just transitioned from unavailable to available
```

internal/datasystem/fdv2_datasystem.go
```diff
@@ -0,0 +1,302 @@
...
+					f.loggers.Warn("Reinitializing data Store from in-memory cache after after data Store outage")
+					if err := f.store.Commit(); err != nil {
+						f.loggers.Error("Failed to reinitialize data Store: %v", err)
+					}
+				}
+			}
```
keelerm84, line 170:
```suggestion
			}
```

internal/datasystem/fdv2_store.go
```diff
@@ -0,0 +1,226 @@
...
+
+// DataAvailability returns the status of the store's data. Defaults means there is no data, Cached means there is
+// data, but it's not guaranteed to be recent, and Refreshed means the data has been refreshed from the server.
+func (s *Store) DataAvailability() DataAvailability {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	return s.availability
+}
+
+// Mirroring returns true data is being mirrored to a persistent store.
```
keelerm84, line 122:
```suggestion
// Mirroring returns true if data is being mirrored to a persistent store.
```

internal/datasystem/fdv2_store.go
```diff
@@ -0,0 +1,226 @@
...
+func (s *Store) mirroring() bool {
+	return s.persistentStore != nil && s.persistentStoreMode == subsystems.StoreModeReadWrite
+}
+
+// nolint:revive // Standard DataSourceUpdateSink method
+func (s *Store) Init(allData []ldstoretypes.Collection, dataStatus datastatus.DataStatus) bool {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	// TXNS-PS: Requirement 1.3.3, must apply updates to in-memory before the persistent Store.
```
keelerm84, line 132: Did we settle on a format for this? Maybe it would be easier if we went with `TXNS-PS(1.3.3): whatever`.

ldclient.go
```diff
@@ -269,7 +291,8 @@ func MakeCustomClient(sdkKey string, config Config, waitFor time.Duration) (*LDC
 		)
 	}
 
-	dataProvider := ldstoreimpl.NewDataStoreEvaluatorDataProvider(store, loggers)
+	// TODO: We can't actually pass STore() here because it wont' swap between the active ones.
```
keelerm84, line 294:
```suggestion
	// TODO: We can't actually pass Store() here because it won't swap between the active ones.
```

internal/datasystem/data_status.go
```diff
@@ -0,0 +1,12 @@
+package datasystem
+
+type DataAvailability string
+
+const (
+	// Defaults means the SDK has no data and will evaluate flags using the application-provided default values.
```
cwaldren-ld, line 6: Maybe this should be a function from timestamp to status.

So a given timestamp is Default, Stale, Fresh.

PR #189, opened 2024-09-16
refactor: hide data source configuration behind dataSystem interface

cwaldren-ld:
This change defines a new `dataSystem` interface, used internally by `LDClient` to implement all functionality related to data store and data source interactions.

The purpose is to be able to swap out the implementation with one suited for FDv2 in a future PR.

### DIFF
--- a/internal/datastore/data_store_eval_impl.go
+++ b/internal/datastore/data_store_eval_impl.go
@@ -9,13 +9,13 @@ import (
 )
 
 type dataStoreEvaluatorDataProviderImpl struct {
-	store   subsystems.DataStore
+	store   subsystems.ReadOnlyStore
 	loggers ldlog.Loggers
 }
 
 // NewDataStoreEvaluatorDataProviderImpl creates the internal implementation of the adapter that connects
 // the Evaluator (from go-server-sdk-evaluation) with the data store.
-func NewDataStoreEvaluatorDataProviderImpl(store subsystems.DataStore, loggers ldlog.Loggers) ldeval.DataProvider {
+func NewDataStoreEvaluatorDataProviderImpl(store subsystems.ReadOnlyStore, loggers ldlog.Loggers) ldeval.DataProvider {
 	return dataStoreEvaluatorDataProviderImpl{store, loggers}
 }
 
--- a/ldclient.go
+++ b/ldclient.go
@@ -67,8 +67,8 @@ const (
 	migrationVarExFuncName = "LDClient.MigrationVariationCtx"
 )
 
-// The dataSystem interface represents the requirements for the client to retrieve data necessary
-// for evaluations, as well as the related status updates related to the data.
+// dataSystem represents the requirements the client has for storing/retrieving/detecting changes related
+// to the SDK's data model.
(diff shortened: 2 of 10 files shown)


internal/datastore/data_store_eval_impl.go
```diff
@@ -9,13 +9,13 @@ import (
 )
 
 type dataStoreEvaluatorDataProviderImpl struct {
```
cwaldren-ld, line 11: This change is because the existing read/write enabled `DataStore` interface isn't actually needed in the evaluator. It only needs to be able to retrieve segments/flags.

subsystems/data_store.go
```diff
@@ -24,21 +26,6 @@ type DataStore interface {
 	// data, and then delete any previously stored items that were not in the input data.
```
cwaldren-ld, line 26: These were moved into `ReadOnlyStore`

ldclient.go
```diff
@@ -583,13 +554,18 @@ func (client *LDClient) SecureModeHash(context ldcontext.Context) string {
 // this does not guarantee that the flags are up to date; if you need to know its status in more detail, use
 // [LDClient.GetDataSourceStatusProvider].
 //
```
cwaldren-ld, line 556: Added this comment because that's what our unit tests expect. Didn't see it documented.

ldclient.go
```diff
@@ -583,13 +554,18 @@ func (client *LDClient) SecureModeHash(context ldcontext.Context) string {
...
+// Additionally, if the client was configured to be offline, this will always return true.
+//
 // If this value is false, it means the client has not yet connected to LaunchDarkly, or has permanently
 // failed. See [MakeClient] for the reasons that this could happen. In this state, feature flag evaluations
 // will always return default values-- unless you are using a database integration and feature flags had
 // already been stored in the database by a successfully connected SDK in the past. You can use
 // [LDClient.GetDataSourceStatusProvider] to get information on errors, or to wait for a successful retry.
 func (client *LDClient) Initialized() bool {
-	return client.dataSource.IsInitialized()
+	if client.offline {
```
cwaldren-ld, line 565: This logic used to work because if the client was in offline mode, then a `NullDataSource` was instantiated (which returned `true` for IsInitialized()). 

Since `dataSystem` doesn't expose `IsInitialized` - just the `DataAvailability` - we need to check that here explicitly. Simpler to understand.

ldclient.go
```diff
@@ -67,6 +67,33 @@ const (
...
+
+	// Stop halts the data system. Should be called when the client is closed to stop any long-running operations.
+	Stop() error
+
+	// Store returns a read-only accessor for the data model.
+	Store() subsystems.ReadOnlyStore
+
+	// DataAvailability indicates what form of data is available.
+	DataAvailability() datasystem.DataAvailability
+}
```
cwaldren-ld, line 91: We could have dataSystem embed the `subsystems.ReadOnlyStore` interface instead of returning it. That way, the data system just is-a store. 🤷‍♂️

keelerm84 (approved):

PR #191, opened 2024-09-17
refactor: streaming/polling processors take DataDestination/StatusReporter interfaces

cwaldren-ld:
This change splits up the monolithic `DataSourceUpdateSink` into two:
1. `DataDestination`: where data goes
2. `StatusReporter`: where statuses go

This makes it easier to mock them for testing, and also represents a cleaner separation of concerns. Finally, these interfaces are slightly different than the `DataSourceUpdateSink` for FDV1 vs FDV2. 

In the new system, we want to be able to do things like passing a payload version in `Init`. 

I've added doc comments that these interfaces are not stable, since they are public.

### DIFF
--- a/internal/sharedtest/mocks/mock_data_destination.go
+++ b/internal/sharedtest/mocks/mock_data_destination.go
@@ -14,7 +14,7 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-// MockDataDestination is a mock implementation of a data destination used tests involving FDv2 data sources.
+// MockDataDestination is a mock implementation of a data destination used by tests involving FDv2 data sources.
 type MockDataDestination struct {
 	DataStore               *CapturingDataStore
 	Statuses                chan interfaces.DataSourceStatus
--- a/internal/sharedtest/mocks/mock_data_destination.go
+++ b/internal/sharedtest/mocks/mock_data_destination.go
@@ -1,0 +1,104 @@
+package mocks
+
+import (
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/launchdarkly/go-server-sdk/v7/interfaces"
+	"github.com/launchdarkly/go-server-sdk/v7/subsystems"
+	"github.com/launchdarkly/go-server-sdk/v7/subsystems/ldstoretypes"
+
+	th "github.com/launchdarkly/go-test-helpers/v3"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// MockDataDestination is a mock implementation of a data destination used tests involving FDv2 data sources.
+type MockDataDestination struct {
+	DataStore               *CapturingDataStore
+	Statuses                chan interfaces.DataSourceStatus
+	dataStoreStatusProvider *mockDataStoreStatusProvider
+	lastStatus              interfaces.DataSourceStatus
+	lock                    sync.Mutex
+}
+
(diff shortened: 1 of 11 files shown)


subsystems/data_source_status_reporter.go
```diff
@@ -0,0 +1,29 @@
...
+	// state, a temporarily disconnected state, or a permanently stopped state.
+	//
+	// If newState is different from the previous state, and/or newError is non-empty, the SDK
+	// will start returning the new status (adding a timestamp for the change) from
+	// DataSourceStatusProvider.GetStatus(), and will trigger status change events to any
+	// registered listeners.
+	//
+	// A special case is that if newState is DataSourceStateInterrupted, but the previous state was
+	// but the previous state was DataSourceStateInitializing, the state will remain at Initializing
+	// because Interrupted is only meaningful after a successful startup.
```
keelerm84:
```suggestion
	// A special case is that if newState is DataSourceStateInterrupted, but the previous state was
	// DataSourceStateInitializing, the state will remain at Initializing because Interrupted is
	// only meaningful after a successful startup.
```

ldcomponents/streaming_data_source_builder_v2_test.go
```diff
@@ -60,9 +60,11 @@ func TestStreamingDataSourceV2Builder(t *testing.T) {
 
 		s := StreamingDataSourceV2()
 
-		dsu := mocks.NewMockDataSourceUpdates(datastore.NewInMemoryDataStore(sharedtest.NewTestLoggers()))
+		dsu := mocks.NewMockDataDestination(datastore.NewInMemoryDataStore(sharedtest.NewTestLoggers()))
```
keelerm84: Rename `dsu` to `dd`?

ldcomponents/streaming_data_source_builder_v2_test.go
```diff
@@ -81,9 +83,11 @@ func TestStreamingDataSourceV2Builder(t *testing.T) {
 
 		s := StreamingDataSourceV2().InitialReconnectDelay(delay).PayloadFilter(filter)
 
-		dsu := mocks.NewMockDataSourceUpdates(datastore.NewInMemoryDataStore(sharedtest.NewTestLoggers()))
+		dsu := mocks.NewMockDataDestination(datastore.NewInMemoryDataStore(sharedtest.NewTestLoggers()))
```
keelerm84: Same here.

internal/sharedtest/mocks/mock_data_destination.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/launchdarkly/go-server-sdk/v7/interfaces"
+	"github.com/launchdarkly/go-server-sdk/v7/subsystems"
+	"github.com/launchdarkly/go-server-sdk/v7/subsystems/ldstoretypes"
+
+	th "github.com/launchdarkly/go-test-helpers/v3"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// MockDataDestination is a mock implementation of a data destination used tests involving FDv2 data sources.
```
keelerm84:
```suggestion
// MockDataDestination is a mock implementation of a data destination used by tests involving FDv2 data sources.
```

keelerm84 (approved):

PR #192, opened 2024-09-17
chore: add fdv2 store & update sources to use fdv2 protocol definitions

cwaldren-ld:
This adds the FDv2 dual-mode store, which supports serving data requests from an in-memory store or from a persistent store (with cache).


internal/datasource/streaming_data_source_events.go
```diff
@@ -16,7 +16,7 @@ var (
 	deleteDataRequiredProperties = []string{"path", "version"} //nolint:gochecknoglobals
```
cwaldren-ld, line 16: This is just a reversion of recent changes that made these public (for use in datasourcev2), but are no longer necessary.

internal/datastore/in_memory_data_store_impl.go
```diff
@@ -4,11 +4,10 @@ import (
 	"sync"
 
 	"github.com/launchdarkly/go-sdk-common/v3/ldlog"
-	"github.com/launchdarkly/go-server-sdk/v7/subsystems"
```
cwaldren-ld: TODO: Update this component's unit tests with tests for SetBasis/ApplyDelta/Dump

internal/datasourcev2/polling_data_source.go
```diff
@@ -18,10 +18,47 @@ const (
 	pollingWillRetryMessage = "will retry at next scheduled poll interval"
 )
 
-// Requester allows PollingProcessor to delegate fetching data to another component.
```
cwaldren-ld, line 21: A bit of new binding code so that the requestor can pass back FDv2 related data.

internal/datasourcev2/polling_data_source.go
```diff
@@ -142,16 +179,23 @@ func (pp *PollingProcessor) Start(closeWhenReady chan<- struct{}) {
...
 		return err
 	}
 
-	// We initialize the store only if the request wasn't cached
-	if !cached {
-		pp.dataDestination.Init(allData, nil)
+	if response.Cached() {
+		return nil
 	}
+
```
cwaldren-ld, line 154: Use the new FDv2 store APIs!

internal/datasourcev2/polling_http_request.go
```diff
@@ -68,42 +71,110 @@ func (r *pollingRequester) BaseURI() string {
...
+				var serverIntent serverIntent
+				err := json.Unmarshal([]byte(event.Data()), &serverIntent)
+				if err != nil {
+					return nil, err
+				} else if len(serverIntent.Payloads) == 0 {
+					return nil, errors.New("server-intent event has no payloads")
+				}
+
+				intent = serverIntent.Payloads[0].Code
+				if intent == "none" {
```
cwaldren-ld: We have a constant for this (or should)

internal/datasourcev2/polling_http_request.go
```diff
@@ -68,42 +71,110 @@ func (r *pollingRequester) BaseURI() string {
...
+				r := jreader.NewReader([]byte(event.Data()))
+				var kind, key string
+				var item ldstoretypes.ItemDescriptor
+				var err error
+				var dataKind datakinds.DataKindInternal
+
+				for obj := r.Object().WithRequiredProperties([]string{versionField, kindField, "key", "object"}); obj.Next(); {
+					switch string(obj.Name()) {
+					case versionField:
+						// version = r.Int()
```
cwaldren-ld: We can take the version now

internal/datasourcev2/polling_http_request.go
```diff
@@ -68,42 +71,110 @@ func (r *pollingRequester) BaseURI() string {
...
+					case kindField:
+						kind = strings.TrimRight(r.String(), "s")
+						dataKind = dataKindFromKind(kind)
+					case "key":
```
cwaldren-ld: We have consts for this now

internal/datasourcev2/polling_http_request.go
```diff
@@ -68,42 +71,110 @@ func (r *pollingRequester) BaseURI() string {
...
+				var err error
+				var dataKind datakinds.DataKindInternal
+
+				for obj := r.Object().WithRequiredProperties([]string{versionField, kindField, "key", "object"}); obj.Next(); {
+					switch string(obj.Name()) {
+					case versionField:
+						// version = r.Int()
+					case kindField:
+						kind = strings.TrimRight(r.String(), "s")
+						dataKind = dataKindFromKind(kind)
```
cwaldren-ld: This needs to use the new kind deserializer

internal/datasourcev2/streaming_data_source.go
```diff
@@ -249,36 +242,29 @@ func (sp *StreamProcessor) consumeStream(stream *es.Stream, closeWhenReady chan<
...
-						sp.loggers.Infof("Unexpected update found in changeset: %s", update)
+
+				//nolint:godox
+				// TODO(SDK-712): If the destination fails to apply the updates should it affect the processor?
+				// This would only happen in the case of a persistent store.
+
+				switch currentChangeSet.intent.Payloads[0].Code {
+				case fdv2proto.IntentTransferFull:
+					{
+						_ = sp.dataDestination.SetBasis(updates, fdv2proto.NoSelector(), true)
```
cwaldren-ld: Shouldn't return errors here. Should handle in data system, then use DataStoreStatus or other mechanism for testing / reporting

internal/datasystem/store.go
```diff
@@ -0,0 +1,258 @@
...
+// stores distinct, it should be much clearer where exactly data is coming from and the behavior should be more
+// predictable at runtime.
+type Store struct {
+	// Source of truth for flag evals (before initialization), or permanently if there are
+	// no initializers/synchronizers configured. Optional; if not defined, only memoryStore is used.
+	persistentStore *persistentStore
+
+	// Source of truth for flag evaluations (once initialized). Before initialization,
+	// the persistentStore may be used if configured.
+	memoryStore *memorystorev2.Store
```
cwaldren-ld, line 55: import alias `memory`?

internal/datasystem/store.go
```diff
@@ -0,0 +1,258 @@
...
+}
+
+// SetBasis sets the basis of the store. Any existing data is discarded. To request data persistence,
+// set persist to true.
+func (s *Store) SetBasis(events []fdv2proto.Event, selector *fdv2proto.Selector, persist bool) error {
+	collections := fdv2proto.ToStorableItems(events)
+	return s.init(collections, selector, persist)
+}
+
+func (s *Store) init(allData []ldstoretypes.Collection, selector *fdv2proto.Selector, persist bool) error {
```
cwaldren-ld: Move this into SetBasis. We don't need this.

internal/datasystem/store.go
```diff
@@ -0,0 +1,258 @@
...
+	return nil
+}
+
+func (s *Store) getActive() subsystems.ReadOnlyStore {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	return s.active
+}
+
+//nolint:revive // Implementation for ReadOnlyStore.
```
cwaldren-ld, line 239: Can we return a pointer to the right store once for each evaluation?

cwaldren-ld:
@keelerm84 , realized why I was confused - the data system config PR has some of the other improvements in the streaming/polling sources (like replacing strings with constants). I tried to only touch the sources enough to get them to interact with the new store. 

I'll have another PR that focuses on cleaning up the sources.

internal/datasourcev2/polling_http_request.go
```diff
@@ -68,42 +71,110 @@ func (r *pollingRequester) BaseURI() string {
...
-		return nil, false, malformedJSONError{err}
+		return nil, malformedJSONError{err}
 	}
 
-	esEvents := make([]es.Event, 0, len(payload.Events))
-	for _, event := range payload.Events {
-		esEvents = append(esEvents, event)
+	parseItem := func(r jreader.Reader, kind datakinds.DataKindInternal) (ldstoretypes.ItemDescriptor, error) {
+		item, err := kind.DeserializeFromJSONReader(&r)
+		return item, err
```
keelerm84: I believe this is equivalent:

```suggestion
		return kind.DeserializeFromJSONReader(&r)
```

Which makes `parseItem` a single line method. At that point, is there value in having it as a a separate function?

internal/datasourcev2/polling_http_request.go
```diff
@@ -68,42 +71,110 @@ func (r *pollingRequester) BaseURI() string {
...
-	} else if len(data) != 1 {
-		return nil, false, malformedJSONError{errors.New("missing expected put event")}
-	}
+	updates := make([]fdv2proto.Event, 0, len(payload.Events))
 
-	putData, ok := data[0].(datasource.PutData)
-	if !ok {
-		return nil, false, malformedJSONError{errors.New("payload is not a PutData")}
-	}
+	var intent fdv2proto.IntentCode
```
keelerm84: Can we consider changing this name to `intentCode` or something? If I see `intent`, I'm going to think the `serverIntent` message more than the code.

internal/datasourcev2/polling_response.go
```diff
@@ -0,0 +1,48 @@
+package datasourcev2
+
+import "github.com/launchdarkly/go-server-sdk/v7/internal/fdv2proto"
+
+// PollingResponse represents the result of a polling request.
+type PollingResponse struct {
+	events   []fdv2proto.Event
+	cached   bool
+	intent   fdv2proto.IntentCode
```
keelerm84, line 9: I would think we would want to rename this as well. But I could be swayed I suppose. 😄

internal/datasourcev2/streaming_data_source.go
```diff
@@ -449,25 +435,17 @@ func (sp *StreamProcessor) GetFilterKey() string {
...
 
 	parseItem := func(r jreader.Reader, kind datakinds.DataKindInternal) (ldstoretypes.ItemDescriptor, error) {
 		item, err := kind.DeserializeFromJSONReader(&r)
 		return item, err
 	}
 
 	for _, event := range events {
-		switch event.Event() {
-		case putEventName:
+		switch fdv2proto.EventName(event.Event()) {
```
keelerm84, line 462: I can't comment on the actual line, but `parseItem` here suffers from the same issue I pointed out above.

internal/datasystem/store.go
```diff
@@ -0,0 +1,252 @@
...
+	"github.com/launchdarkly/go-server-sdk/v7/internal/fdv2proto"
+
+	"github.com/launchdarkly/go-sdk-common/v3/ldlog"
+	"github.com/launchdarkly/go-server-sdk/v7/interfaces"
+	"github.com/launchdarkly/go-server-sdk/v7/subsystems"
+	"github.com/launchdarkly/go-server-sdk/v7/subsystems/ldstoretypes"
+)
+
+// Store is a dual-mode persistent/in-memory store that data queries from the evaluation
+// algorithm.
```
keelerm84: `that data queries from` doesn't make sense to me. Can we re-word that?

internal/datasystem/store.go
```diff
@@ -0,0 +1,252 @@
...
+//
+// At any given moment one of two stores is active: in-memory, or persistent. This doesn't preclude a caller
+// from holding on to a reference to the persistent store even when we swap to the in-memory store.
+//
+// Once the in-memory store has data (either from initializers or a synchronizer), the persistent
+// store is no longer read from. From that point forward, calls to Get will serve data from the memory
+// store.
+//
+// One motivation behind using persistent stores is to offer a way to immediately start evaluating
+// flags before a connection is made to LD (or even in the moment before an initializer has run.)
```
keelerm84: Sorry for the nitpick.

```suggestion
// flags before a connection is made to LD (or even in the moment before an initializer has run).
```

internal/fdv2proto/event_to_storable_item.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/launchdarkly/go-server-sdk/v7/internal/datakinds"
+	"github.com/launchdarkly/go-server-sdk/v7/subsystems/ldstoretypes"
+)
+
+// ToStorableItems converts a list of FDv2 events into a slice of collections, which is suitable for
+// passing into a store.
+func ToStorableItems(events []Event) []ldstoretypes.Collection {
+	flagCollection := ldstoretypes.Collection{
+		Kind:  datakinds.Features,
+		Items: make([]ldstoretypes.KeyedItemDescriptor, 0),
```
keelerm84, line 13: Do you think it's worth reserving some capacity here to prevent the additional allocations?

keelerm84 (approved):

internal/datasourcev2/streaming_data_source.go
cwaldren-ld: It turns out this whole bit of logic (the event parsing) doesn't work properly due to the streaming JSON parser. I'm going to have to refactor this to more closely match how the existing datasource (V1) parsers work - which is two passes. 

The reason is because we might visit the "object" field _before_ we've visited the kind field.

internal/datasystem/store.go
cwaldren-ld: Ah, was missing a word. 👍

internal/fdv2proto/event_to_storable_item.go
cwaldren-ld: We don't know the size up front, although we could calculate it.  I think I'd rather let Go's allocation strategy do its thing and if it's a problem we can fix it.

PR #193, opened 2024-09-24
chore: introduce memorystorev2

cwaldren-ld:
**Requirements**

- [x] I have added test coverage for new or changed functionality

This adds a `memorystorev2` package containing an alternative Memory Store implementation designed specifically for FDv2 usage.

It is similar to the existing memory store, with obsolete methods removed and new methods added - but the core `upsert/init` logic remains the same. 

I've converted the existing unit tests and benchmarks to the new store. 

Augment the existing memory store with FDv2 functionality. I did this at first, and although it worked, it wasn't clean. It meant the store needed to retain obsolete functions, and it wasn't clear how the FDv2 methods interacted with FDv1 methods. Example: `SetBasis` vs `Init`, seems redundant. 

**Additional context**

This component is necessary for the dual-mode persistent/memory store PR.

### DIFF
--- a/internal/memorystorev2/memory_store.go
+++ b/internal/memorystorev2/memory_store.go
@@ -21,8 +21,8 @@ import (
 // atomically. The idea is that there's never a moment when the state of the store could be inconsistent
 // with regard to the authoritative LaunchDarkly SaaS.
 type Store struct {
-	allData       map[ldstoretypes.DataKind]map[string]ldstoretypes.ItemDescriptor
-	isInitialized bool
+	data        map[ldstoretypes.DataKind]map[string]ldstoretypes.ItemDescriptor
+	initialized bool
 	sync.RWMutex
 	loggers ldlog.Loggers
 }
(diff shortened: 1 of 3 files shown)


internal/memorystorev2/memory_store.go
```diff
@@ -0,0 +1,171 @@
...
+// It accepts updates in batches - for instance, flag A was upserted while segment B was deleted -
+// such that the contents of the store are consistent with a single payload version at any given time.
+//
+// The terminology used is "basis" and "deltas". First, the store's basis is set. This is this initial
+// data, upon which subsequent deltas will be applied. Whenever the basis is set, any existing data
+// is discarded.
+//
+// Deltas are then applied to the store. A single delta update transforms the contents of the store
+// atomically. The idea is that there's never a moment when the state of the store could be inconsistent
+// with regard to the authoritative LaunchDarkly SaaS.
```
keelerm84: That might be interpreted in a way that is over promising. The SDKs can be out of sync with the SaaS system, and so people may think never being inconsistent might mean never being out of sync. Maybe we should try to clarify that language?

keelerm84 (approved):

PR #198, opened 2024-09-26
refactor: more fdv2 protocol types + usage in data sources

cwaldren-ld:
This fully converts the datasources to use `fdv2` protocol domain objects.


internal/datasourcev2/streaming_data_source.go
```diff
@@ -438,81 +453,74 @@ func (sp *StreamProcessor) GetFilterKey() string {
...
-					kind := r.String()
-					dataKind = dataKindFromKind(kind)
-					if dataKind == nil {
-						//nolint: godox
-						// TODO: We are skipping here without showing a warning. Need to address that later.
-						continue
-					}
+					kind = fdv2proto.ObjectKind(r.String())
+					//nolint:godox
+					// TODO: An unrecognized kind should be ignored for forwards compat; the question is,
```
keelerm84, line 485: That's a good question. Generally I like to handle error cases as close to the source of the error as possible. In this case, I think I can see the argument for saying it isn't an error in the data source. It's only an error once you try to apply it to something that doesn't understand it. 

When we go to update our SDKs in the future to support a third kind, would I expect to look at the data source or the data store? For that reason alone, I think we can let the store handle this.

internal/fdv2proto/payloads.go
```diff
@@ -0,0 +1,18 @@
+package fdv2proto
+
+type Payload struct {
+	// The id here doesn't seem to match the state that is included in the
+	// Payload transferred object.
+
+	// It would be nice if we had the same value available in both so we could
+	// use that as the key consistently throughout the the process.
+	ID     string     `json:"id"`
+	Target int        `json:"target"`
```
keelerm84, line 10: As it currently stands, I think we can probably ignore these fields completely right? The SDK is going to operate under the assumption we will receive a single payload, and that payload is being applied to the entire SDK. So I don't think we actually care that much about the ID and Target fields.

keelerm84 (approved):

internal/datasourcev2/streaming_data_source.go
cwaldren-ld: That makes sense to me.

PR #199, opened 2024-10-11
chore: Expand contract tests to support persistent data stores

keelerm84:


keelerm84:
Note that this is being merged into a feature branch. There is still ongoing development of this feature, so I don't want to merge it into the mainline yet. However, it is good to start getting eyes on this part too.

.github/workflows/common_ci.yml
```diff
@@ -29,7 +29,21 @@ jobs:
...
+
+    services:
+      redis:
+        image: redis
+        ports:
+            - 6379:6379
+      dynamodb:
+        image: amazon/dynamodb-local
+        ports:
+            - 8000:8000
```
cwaldren-ld, line 42: Kinda surprised they "took" port 8000 for dynamodb.

testservice/sdk_client_entity.go
```diff
@@ -384,6 +394,59 @@ func makeSDKConfig(config servicedef.SDKConfigParams, sdkLog ldlog.Loggers) ld.C
...
+			if config.PersistentDataStore.Store.Prefix != nil {
+				dsBuilder.Prefix(*config.PersistentDataStore.Store.Prefix)
+			}
+			builder = ldcomponents.PersistentDataStore(dsBuilder)
+		case servicedef.DynamoDB:
+			cfg, err := awsconfig.LoadDefaultConfig(context.Background(),
+				awsconfig.WithRegion("us-east-1"),
+				awsconfig.WithCredentialsProvider(credentials.NewStaticCredentialsProvider("dummy", "dummy", "dummy")),
+			)
+			if err != nil {
```
cwaldren-ld, line 425: The function that calls `makeSDKConfig` is capable of returning an error, so it would be good to change the signature here so we can bubble this back up

testservice/sdk_client_entity.go
```diff
@@ -384,6 +394,59 @@ func makeSDKConfig(config servicedef.SDKConfigParams, sdkLog ldlog.Loggers) ld.C
...
+				o.EndpointResolver = dynamodb.EndpointResolverFromURL(config.PersistentDataStore.Store.DSN)
+			})
+
+			dsBuilder := lddynamodb.DataStore("sdk-contract-tests").DynamoClient(svc)
+			if config.PersistentDataStore.Store.Prefix != nil {
+				dsBuilder.Prefix(*config.PersistentDataStore.Store.Prefix)
+			}
+
+			builder = ldcomponents.PersistentDataStore(dsBuilder)
+		default: // TODO: We should probably fail here.
```
cwaldren-ld: This could also return an error

testservice/service.go
```diff
@@ -246,7 +249,7 @@ func main() {
 	loggers := ldlog.NewDefaultLoggers()
 	loggers.SetMinLevel(logLevelFromName(os.Getenv("LD_LOG_LEVEL")))
 
-	port := "8000"
+	port := "10000"
```
cwaldren-ld, line 252: Was changing the dynamo port a no-go, rather than changing the test service's port?

testservice/servicedef/service_params.go
```diff
@@ -9,22 +9,25 @@ const (
...
+	CapabilityInlineContext               = "inline-context"
+	CapabilityAnonymousRedaction          = "anonymous-redaction"
+	CapabilityEvaluationHooks             = "evaluation-hooks"
+	CapabilityOmitAnonymousContexts       = "omit-anonymous-contexts"
+	CapabilityEventGzip                   = "event-gzip"
+	CapabilityOptionalEventGzip           = "optional-event-gzip"
+	CapabilityClientPrereqEvents          = "client-prereq-events"
+	CapabilityPersistentDataStoreRedis    = "persistent-data-store-redis"
+	CapabilityPersistentDataStoreConsul   = "persistent-data-store-consul"
+	CapabilityPersistentDataStoreDynamoDB = "persistent-data-store-dynamodb"
```
cwaldren-ld, line 30: Not for this PR: I think we really don't need constants for these things.

cwaldren-ld (approved):

cwaldren-ld:
Awesome work!

testservice/sdk_client_entity.go
keelerm84: Done.

testservice/sdk_client_entity.go
keelerm84: Done.

testservice/service.go
keelerm84: I could have changed the dynamodb port. But since every other SDK is going to have to override it (and then each of us will when doing local work), I thought it might be better to try and get those ports on the standard ones, and then maybe we can align on an SDK test harness port.

PR #201, opened 2024-10-14
feat: add support for client-side prerequisite events

cwaldren-ld:
This PR updates `AllFlagsState` to track prerequisite evaluations. 

This didn't require modifying the eval module, as it already exposes a `PrerequisiteEventRecorder` interface with the necessary info.

The `AllFlagsState` public API allows for fetching a flag's details (`FlagState`) from the top-level `AllFlags` object. This struct now has prerequisite information exposed. 

Additionally when the `AllFlags` is marshaled to JSON, it will now contain the prerequisite info for each flag.

### DIFF
--- a/interfaces/flagstate/flags_state.go
+++ b/interfaces/flagstate/flags_state.go
@@ -67,6 +67,10 @@ type FlagState struct {
 	// OmitDetails is true if, based on the options passed to AllFlagsState and the flag state, some of the
 	// metadata can be left out of the JSON representation.
 	OmitDetails bool
+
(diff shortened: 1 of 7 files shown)


interfaces/flagstate/flags_state.go
```diff
@@ -67,6 +67,10 @@ type FlagState struct {
 	// OmitDetails is true if, based on the options passed to AllFlagsState and the flag state, some of the
 	// metadata can be left out of the JSON representation.
 	OmitDetails bool
+
```
cwaldren-ld, line 70: This is a new exposed field on the public API, so this PR needs to be a feat.

kinyoklion (approved):

PR #207, opened 2024-11-26
refactor: add AI client and unit tests

cwaldren-ld:
This implements the bulk of the AI client, consisting of the Config builder and Tracker structs. 

I've also added many interpolation tests, config builder tests, and tracking tests.


ldai/client.go
```diff
@@ -1,12 +1,139 @@
...
+	}, nil
+}
+
+func (c *Client) Config(
+	key string,
+	context ldcontext.Context,
+	defaultValue Config,
+	variables map[string]interface{},
+) (Config, *Tracker) {
+	variation, err := c.sdk.JSONVariation(key, context, ldvalue.Null())
```
keelerm84: Okay, so this is something we should talk about because I thought about doing this same thing.

You are passing in null here. In the python implementation, we are passing the default value through. In certain paths, you and I are emitting different shaped events since the default values are included in there.

We should probably figure out which one it should be.

ldai/client.go
```diff
@@ -1,12 +1,139 @@
...
+	for _, ctx := range context.GetAllIndividualContexts(nil) {
+		attributes[string(ctx.Kind())] = addSingleKindContextAttributes(ctx)
+	}
+
+	return attributes
+}
+
+func addSingleKindContextAttributes(context ldcontext.Context) map[string]interface{} {
+	attributes := map[string]interface{}{
+		"kind":      context.Kind(),
```
keelerm84: Hmmm. This is a really slight nitpick, and it may never matter, but in the multi-kind case, I don't think we should include this here for consistency with other SDKs.

ldai/client.go
cwaldren-ld: Ah, yea, I had left for the day before pushing up my latest change which passes in the marshalled default value. I think that is the most correct way to do this with the least amount of cleverness.

ldai/client.go
cwaldren-ld: I did add `kind` in .NET as well 🤷‍♂️ 

Can remove. But it seems like if you are expecting to be able to access `kind`, you should be able to - even if it's `multi`.

keelerm84: I agree you should be able to access the `kind` as 'multi'. My concern is that this allows for `ldctx.org.kind`, which doesn't match our context schema.

cwaldren-ld: Got it. Fixed.

ldai/client.go
```diff
@@ -42,38 +48,47 @@ func New(sdk ServerSDK) (*Client, error) {
...
+func (c *Client) logConfigWarning(key string, format string, args ...interface{}) {
+	prefix := "AI config '" + key + "': "
+	c.logger.Warnf(prefix+format, args...)
+}
+
+// Config evaluates an AI config named by a given key for the given context.
+//
+// The config's messages will undergo Mustache template interpolation using the provided variables, which may be
+// nil. If the config cannot be evaluated or LaunchDarkly is unreachable, the default value is returned. Note that
+// the messages in the default will not undergo template interpolation.
```
keelerm84, line 63: > Note that the messages in the default will not undergo template interpolation.

That's not the case for python, and I'm not sure it's the case for JS either. Might be a case where the spec is under-specified. @kinyoklion thoughts?

keelerm84 (approved):

ldai/client.go
cwaldren-ld: How does it work if the template interpolation fails?

kinyoklion: We just have not handled this in python and JS yet. We did it this way in .Net.

I am still not sure if it is a great solution, but we do need to choose some solution.

keelerm84: > How does it work if the template interpolation fails?
> > If the config cannot be evaluated or LaunchDarkly is unreachable

If the LDClient is offline, we will return the default value from the variation method. In python, that is the default config value as a dictionary. So we get back something in the AI client that looks like the right shape, and so we can still do the templating for ldctx at least.

That seems like something that would be useful for customers.

kinyoklion: If your template string is invalid in your default value the interpolation can still fail. Which is a hangup itself.

keelerm84: Yes, templating related failures are a problem we need to surface as well. I just want to make sure we're handling at least the case where we could do templating for a default value, but aren't.

cwaldren-ld: Closing the loop, we should discuss more, but I won't consider this blocking for getting out the alpha.

PR #211, opened 2024-12-05
fix: pass parsed versionKey to Tracker

cwaldren-ld:
The AI client wasn't passing the parsed `_ldMeta.versionKey` into newly created `Tracker`s. The tests didn't catch it because the default value of strings are `""`, and that's what the tests asserted against.

I've updated the tests to pass in a non-default string to make sure it is being plumbed through.

### DIFF
--- a/ldai/client.go
+++ b/ldai/client.go
@@ -76,13 +76,13 @@ func (c *Client) Config(
 	// empty object.)
 	if result.Type() != ldvalue.ObjectType {
 		c.logConfigWarning(key, "unmarshalling failed, expected JSON object but got %s", result.Type().String())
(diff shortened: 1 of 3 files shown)


ldai/client.go
```diff
@@ -76,13 +76,13 @@ func (c *Client) Config(
 	// empty object.)
 	if result.Type() != ldvalue.ObjectType {
 		c.logConfigWarning(key, "unmarshalling failed, expected JSON object but got %s", result.Type().String())
-		return defaultValue, newTracker(key, c.sdk, &defaultValue, context, c.logger)
```
cwaldren-ld, line 79: I think we agreed on "empty string for version key in default value".

kinyoklion (approved):

PR #223, opened 2024-12-11
refactor: add fallback/recovery additions [prototype]

cwaldren-ld:
This PR attempts to demonstrate how fallback/recovery conditions might be implemented in FDv2.

The main idea is that the SDK can fallback to a secondary synchronizer, and then recover to the primary synchronizer at a later date.

Some challenges I've identified when implementing this PR:
1. Conditions require accurate status information to be evaluated correctly. For example, a condition that executes "when the data source is interrupted for 5 seconds" needs to know that the status is "interrupted" and that it has been interrupted for 5 seconds. So these need to be accessible to the condition, and it's important they be set correctly. 
2. We'll need some way of specifying the conditions in our configuration system. We can start off with simple conditions like:   `fallback { unhealthy_duration: duration } / recover { healthy_duration: duration, unhealthy_duration: duration}`, and specify some reasonable defaults. 
3. The current data sources aren't setup for being restarted. That is, they expect to be started and then closed - they aren't designed specifically to have `Sync(..)` called multiple times. This means we can't be sure they are safe for the current way I'm using them in the PR. 
    - A better design would be having a `Run(ctx context.Context, selector)` method that is synchronous from the perspective of the Data System algorithm. When the function returns, we know for sure that the data source is closed. Then we can simply call `Run` again. Another option is to keep the existing `Sync/Close` pattern, but make sure it is safe for multiple usage (that is, once `Close` is called, it is safe to call `Sync` again.)

4. The existing pattern of `closeWhenReady` channel is difficult to use in a world with > 1 data source (and ones that can be restarted.) I've added some thoughts about that in the code comments.

### DIFF
--- a/internal/datasourcev2/polling_http_request.go
+++ b/internal/datasourcev2/polling_http_request.go
@@ -72,12 +72,9 @@ func (r *pollingRequester) Request() (*fdv2proto.ChangeSet, error) {
 		r.loggers.Debug("Polling LaunchDarkly for feature flag updates")
 	}
 
-	body, cached, err := r.makeRequest(endpoints.PollingRequestPath)
+	body, _, err := r.makeRequest(endpoints.PollingRequestPath)
 	if err != nil {
 		return nil, err
-	}
-	if cached {
-		return fdv2proto.NewChangeSetBuilder().NoChanges(), nil
 	}
 
 	var payload fdv2proto.PollingPayload
--- a/internal/datasourcev2/streaming_data_source.go
+++ b/internal/datasourcev2/streaming_data_source.go
@@ -287,6 +287,7 @@ func (sp *StreamProcessor) consumeStream(stream *es.Stream, closeWhenReady chan<
 				sp.setInitializedAndNotifyClient(true, closeWhenReady)
 
 			default:
+				processedEvent = false
 				sp.loggers.Infof("Unexpected event found in stream: %s", event.Event())
 			}
 
--- a/internal/datasystem/fdv2_datasystem.go
+++ b/internal/datasystem/fdv2_datasystem.go
@@ -114,6 +114,7 @@ func NewFDv2(disabled bool, cfgBuilder subsystems.ComponentConfigurer[subsystems
 	fdv2.disabled = disabled
 	fdv2.fallbackCond = func() bool {
 		status := fdv2.getStatus()
+		fdv2.loggers.Debugf("Status: %s", status.String())
 		interruptedAtRuntime := status.State == interfaces.DataSourceStateInterrupted && time.Since(status.StateSince) > 1*time.Minute
 		cannotInitialize := status.State == interfaces.DataSourceStateInitializing && time.Since(status.StateSince) > 30*time.Second
 		return interruptedAtRuntime || cannotInitialize
@@ -284,6 +285,7 @@ func (f *FDv2) evaluateCond(ctx context.Context, cond func() bool) error {
 			if cond() {
 				return nil
 			}
+			f.loggers.Debugf("Condition check succeeded, continue with current synchronizer")
 		}
 	}
 }
--- a/internal/datasystem/fdv2_datasystem.go
+++ b/internal/datasystem/fdv2_datasystem.go
@@ -6,8 +6,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/launchdarkly/go-server-sdk/v7/internal/fdv2proto"
-
 	"github.com/launchdarkly/go-sdk-common/v3/ldlog"
 	"github.com/launchdarkly/go-server-sdk/v7/interfaces"
 	"github.com/launchdarkly/go-server-sdk/v7/internal"
@@ -72,6 +70,9 @@ type FDv2 struct {
 	// Protects status.
 	mu     sync.Mutex
 	status interfaces.DataSourceStatus
+
+	fallbackCond func() bool
+	recoveryCond func() bool
 }
 
 // NewFDv2 creates a new instance of the FDv2 data system. The first argument indicates if the system is enabled or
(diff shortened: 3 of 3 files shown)


internal/datasourcev2/polling_http_request.go
```diff
@@ -72,13 +72,10 @@ func (r *pollingRequester) Request() (*fdv2proto.ChangeSet, error) {
 		r.loggers.Debug("Polling LaunchDarkly for feature flag updates")
 	}
 
-	body, cached, err := r.makeRequest(endpoints.PollingRequestPath)
```
cwaldren-ld, line 75: This is a hack because our current data sources are started/stopped rather than being re-instantiated. The issue is - if we had an error response (like the payload was malformed or just an HTTP error), we'd get back an empty changeset (`.NoChanges()`) with no error (`nil`).

This means if the previous Data Source Status was something like `VALID` due to a previous synchronizer, then we start up polling and it gets the same response as it did last time, we wouldn't update the state to `INTERRUPTED` based on this error.

If instead the data source was re-instantiated, then there would be no "previous state" for this new run of the data source. So we'd get the error, update the status, and then get the error again (cached) and not update the status - but that'd be correct from the data system's point of view, since nothing has changed.

internal/datasourcev2/streaming_data_source.go
```diff
@@ -287,6 +287,7 @@ func (sp *StreamProcessor) consumeStream(stream *es.Stream, closeWhenReady chan<
 				sp.setInitializedAndNotifyClient(true, closeWhenReady)
 
 			default:
+				processedEvent = false
```
cwaldren-ld, line 290: This seems incorrect in both the fdv1 and fdv2 sources. If we get an unrecognized event, then we don't want to set the data source state to valid. That would clear any existing error.

internal/datasystem/fdv2_datasystem.go
```diff
@@ -111,6 +112,19 @@ func NewFDv2(disabled bool, cfgBuilder subsystems.ComponentConfigurer[subsystems
 	fdv2.primarySync = cfg.Synchronizers.Primary
 	fdv2.secondarySync = cfg.Synchronizers.Secondary
 	fdv2.disabled = disabled
```
cwaldren-ld, line 114: The power of conditions is the chaining I've shown here. We can have an arbitrary number of conditions and hook them up, which I can see being useful in the (far) future. For now, it'd probably be fine to define some preset conditions with a couple of knobs.

internal/datasystem/fdv2_datasystem.go
```diff
@@ -111,6 +112,19 @@ func NewFDv2(disabled bool, cfgBuilder subsystems.ComponentConfigurer[subsystems
...
+	fdv2.fallbackCond = func(status interfaces.DataSourceStatus) bool {
+		interruptedAtRuntime := status.State == interfaces.DataSourceStateInterrupted && time.Since(status.StateSince) > 1*time.Minute
+		cannotInitialize := status.State == interfaces.DataSourceStateInitializing && time.Since(status.StateSince) > 10*time.Second
+		healthyForTooLong := status.State == interfaces.DataSourceStateValid && time.Since(status.StateSince) > 30*time.Second
```
cwaldren-ld, line 118: `healthyForTooLong` is an interesting one. We want this in the `recoveryCond` in order to prevent the SDK from using the secondary for too long - presumably we want to switch back to the primary because it is more efficient/better for [reasons]. 

I put it in the `fallbackCond` to cause a flip-flop pattern for demo purposes. We probably wouldn't actually want that condition in there. Although, it could be useful in a chaos monkey sense - every so often, check that your backup is functioning.

internal/datasystem/fdv2_datasystem.go
```diff
@@ -302,13 +383,17 @@ func (f *FDv2) Offline() bool {
 }
 
 //nolint:revive // DataSourceStatusReporter method.
-func (f *FDv2) UpdateStatus(status interfaces.DataSourceState, err interfaces.DataSourceErrorInfo) {
+func (f *FDv2) UpdateStatus(state interfaces.DataSourceState, err interfaces.DataSourceErrorInfo) {
```
cwaldren-ld, line 386: This function equivalent in fdv1 is here: https://github.com/launchdarkly/go-server-sdk/blob/v7/internal/datasource/data_source_update_sink_impl.go#L157

The minimal implementation I wrote here is for demo purposes. We may need to adopt the other one to be backwards compatible.

internal/datasourcev2/streaming_data_source.go
keelerm84: I'm wondering if this shouldn't be a tri-bool and then we have a little different handling, where:

1. We only set it to true if we have processed a valid event.
2. Failure to decode an expected event fails would set this to false.
3. An unknown event type would be ignored, with no change to this value, as a sort of forward compatible guard against new event types we want to add.
4. We only execute line 294 if it is true, and avoid the false and new third (null?) state.

internal/datasystem/fdv2_datasystem.go
```diff
@@ -205,35 +222,99 @@ func (f *FDv2) runInitializers(ctx context.Context, closeWhenReady chan struct{}
...
+// The Go SDK is architected in such a way that a "closeWhenReady" channel is passed into the Data System by means
+// of MakeClient. MakeClient is able to block the user until the client reaches a terminal state in regard to
+// its initialization status - either initialized with data, or run out of time initializing, or some fatal error.
+//
+// By closing this channel, the SDK's data sources indicate that MakeClient should unblock.
+//
+// In the FDv2 world, we have the possibility that a synchronizer fails or we fall back to a secondary synchronizer.
+// Perhaps we've already closed the channel, and now a new synchronizer is attempting to do the same.
+//
+// In that case, we need to guarantee that the channel is closed only once. To do this, we "wrap" channel that is passed
```
keelerm84, line 237:
```suggestion
// In that case, we need to guarantee that the channel is closed only once. To do this, we "wrap" the channel that is passed
```

internal/datasystem/fdv2_datasystem.go
```diff
@@ -205,35 +222,99 @@ func (f *FDv2) runInitializers(ctx context.Context, closeWhenReady chan struct{}
...
+			case <-ctx.Done():
+				return
+			default:
+			}
+		}
+	})
+}
+
+func (f *FDv2) evaluateCond(ctx context.Context, cond func(status interfaces.DataSourceStatus) bool) error {
+	ticker := time.NewTicker(10 * time.Second)
```
keelerm84, line 305: This hard coded 10 second timer is what limits the resolution of the fallback / recovery conditions right?

cwaldren-ld: Correct. I could see an alternative of making this event-triggered, where we'd have "timer event" and "data source status event" (and anything else that can be used as a condition.)

But then we'd need to hold a map of timers, hook into the data source status broadcasters.. it just doesn't seem worth the complexity when compared to a predictable "tick" that polls whatever data is needed.

internal/datasourcev2/streaming_data_source.go
cwaldren-ld: Seems like a good idea. 

```
enum ProcessingState {
    EVENT_DECODED,
    EVENT_IGNORED,
    EVENT_MALFORMED
}
```
or similar.

keelerm84:
Done in https://github.com/launchdarkly/go-server-sdk/pull/242 instead. Feedback from this was adopted there.

So long @cwaldren-ld, and thanks for all the 🐟

PR #227, opened 2024-12-18
fix: Send updated selector on streaming reconnect

keelerm84:
BEGIN_COMMIT_OVERRIDE
chore: Send updated selector on streaming reconnect
END_COMMIT_OVERRIDE


keelerm84:
This relies on a pending eventsource change, but at least this gets the PR up and ready for review.

internal/datasourcev2/streaming_data_source.go
```diff
@@ -304,11 +304,8 @@ func (sp *StreamProcessor) consumeStream(stream *es.Stream, closeWhenReady chan<
 	}
 }
 
-func (sp *StreamProcessor) subscribe(closeWhenReady chan<- struct{}, selector fdv2proto.Selector) {
+func (sp *StreamProcessor) subscribe(closeWhenReady chan<- struct{}, _ fdv2proto.Selector) {
```
tanderson-ld, line 307: Do you think this param will ever be used agian?

subsystems/data_destination.go
```diff
@@ -11,6 +11,9 @@ import (
 // Do not use it.
 // You have been warned.
 type DataDestination interface {
+	// Selector returns the last known selector for the data store.
```
tanderson-ld: Consider adding that the selector may be unknown.  Is that the `fdv2proto.NoSelector()`?

subsystems/data_destination.go
```diff
@@ -11,6 +11,9 @@ import (
 // Do not use it.
```
tanderson-ld, line 13: Update comment to remove use of "sink".

internal/datasourcev2/streaming_data_source.go
keelerm84: Given our latest discussion, I wouldn't think so. But I don't want to remove it in this PR just yet as it's going to balloon up. How about I remove it in a separate PR?

subsystems/data_destination.go
keelerm84: Done.

subsystems/data_destination.go
keelerm84: Do you have a recommendation for different wording? I'm not sure what I would say differently here.

internal/datasourcev2/streaming_data_source.go
keelerm84: https://github.com/launchdarkly/go-server-sdk/pull/228

subsystems/data_destination.go
keelerm84: Ryan had a good suggestion. How does that sound?

kinyoklion (approved):

tanderson-ld (approved):

tanderson-ld (approved):

PR #232, opened 2024-12-23
chore: Fail initialization if data system doesn't archive target availability

keelerm84:


tanderson-ld (approved):

ldclient.go
```diff
@@ -342,7 +342,7 @@ func MakeCustomClient(sdkKey string, config Config, waitFor time.Duration) (*LDC
 		for {
 			select {
 			case <-closeWhenReady:
-				if client.dataSystem.DataAvailability() != datasystem.Refreshed {
+				if client.dataSystem.DataAvailability() != datasystem.Defaults {
```
keelerm84: (self-reminder): I can't make this change as it would be a modification of existing behavior.

tanderson-ld (approved):
Assuming the case of getting Refreshed when target is Cached is impossible.

PR #234, opened 2025-01-07
chore: Plumb context.Context throughout polling datasourcev2

keelerm84:


internal/datasourcev2/polling_data_source.go
```diff
@@ -93,6 +91,8 @@ func (pp *PollingProcessor) Fetch(_ context.Context) (*subsystems.Basis, error)
 func (pp *PollingProcessor) Sync(closeWhenReady chan<- struct{}) {
 	pp.loggers.Infof("Starting LaunchDarkly polling with interval: %+v", pp.pollInterval)
 
+	ctx := context.TODO()
```
kinyoklion, line 99: Maybe we should comment this, as I assume it isn't the ideal.

keelerm84: Done.

kinyoklion (approved):

PR #235, opened 2025-01-09
chore: Remove TODO placeholder for implicit change intent

keelerm84:


internal/fdv2proto/changeset.go
```diff
@@ -93,11 +93,11 @@ func (c *ChangeSetBuilder) Finish(selector Selector) (*ChangeSet, error) {
 		changes:    c.changes,
 	}
 	c.changes = nil
+
+	// Once a full transfer has been processed, all future changes should be
+	// assumed to be changes. Flag delivery can override this behavior by
+	// sending a new ServerIntent to any connected stream.
 	if c.intent.Payload.Code == IntentTransferFull {
-		//nolint:godox
-		// TODO(SDK-931): We have an awkward situation where we don't get a new intent after receiving a payload
```
keelerm84, line 98: I actually don't see this as that awkward personally.

After the initial payload is received, we assume all future modifications will be updates instead of full replacements. However, we do allow for the possibility of receiving a new server-intent in the future which would allow FD to optimize with a xfer-full instead of a xfer-changes.

It should also be noted the SDK test harness' tests already cover both situations, so we are covered on that side too.

tanderson-ld (approved):

PR #241, opened 2025-01-22
feat: add timeToFirstToken to Tracker

moribellamy:
BEGIN_COMMIT_OVERRIDE
feat: Add timeToFirstToken to Tracker
feat: Drop Set method on Metric interface.
END_COMMIT_OVERRIDE

### DIFF
--- a/ldai/tracker.go
+++ b/ldai/tracker.go
@@ -36,17 +36,17 @@ type TokenUsage struct {
 	Output int
 }
 
+// Set returns true if any of the fields are non-zero.
+func (t TokenUsage) Set() bool {
+	return t.Total > 0 || t.Input > 0 || t.Output > 0
+}
+
 // Metrics represents the metrics returned by a model provider for a specific request.
 type Metrics struct {
 	// Latency is the latency of the request.
 	Latency time.Duration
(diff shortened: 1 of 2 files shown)


ldai/tracker.go
```diff
@@ -43,11 +45,13 @@ func (t TokenUsage) Set() bool {
...
 	// Latency is the latency of the request.
 	Latency time.Duration
+	// TimeToFirstToken is the time to the first token of the streamed response.
+	TimeToFirstToken time.Duration
 }
 
 // Set returns true if the latency is non-zero.
 func (m Metrics) Set() bool {
-	return m.Latency != 0
+	return m.Latency != 0 || m.TimeToFirstToken != 0
```
keelerm84: Given the refactoring below, it doesn't appear like this method is still in use anywhere.

We could remove this method, and then this would be a breaking change (which is fine since we're pre-v1.

Or if we really don't want to remove it, we should at least update the documentation to match the current behavior.

keelerm84 (changes requested):

ldai/tracker.go
moribellamy: i only kept it for fear of breaking change. i'll remove it!

keelerm84 (approved):

PR #256, opened 2025-04-15
fix: Handle shutting down data system when all synchronizers permanently fail

keelerm84:
BEGIN_COMMIT_OVERRIDE
chore: Handle shutting down data system when all synchronizers permanently fail
END_COMMIT_OVERRIDE

### DIFF
--- a/internal/datasourcev2/polling_data_source.go
+++ b/internal/datasourcev2/polling_data_source.go
@@ -2,7 +2,6 @@ package datasourcev2
 
 import (
 	"context"
-	"sync"
 	"sync/atomic"
 	"time"
 
@@ -37,9 +36,9 @@ type PollingProcessor struct {
 	requester       PollingRequester
 	pollInterval    time.Duration
 	loggers         ldlog.Loggers
-	isInitialized   *atomic.Bool
+	isInitialized   atomic.Bool
+	isClosed        atomic.Bool
 	quit            chan struct{}
-	closeOnce       sync.Once
 }
 
 // NewPollingProcessor creates the internal implementation of the polling data source.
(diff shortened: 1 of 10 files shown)


internal/datasourcev2/polling_data_source.go
```diff
@@ -174,12 +178,18 @@ func (pp *PollingProcessor) poll(ctx context.Context) error {
...
 		pp.dataDestination.ApplyDelta(changeSet.Changes(), changeSet.Selector(), true)
-	case fdv2proto.IntentNone:
-		{
-			// no-op, we are already up-to-date.
+		statusChan <- interfaces.DataSourceStatus{
+			State:      interfaces.DataSourceStateValid,
+			StateSince: time.Now(),
 		}
+	case fdv2proto.IntentNone:
+		// no-op, we are already up-to-date
```
tanderson-ld, line 181: I think this should report valid.  The transfer none is positive confirmation in our application layer that the datasource is receiving state, in this case, no transfer of state is required.

internal/datasystem/fdv2_datasystem.go
```diff
@@ -317,18 +316,44 @@ func (f *FDv2) runSynchronizers(ctx context.Context, closeWhenReady chan struct{
...
 		select {
 		case <-ctx.Done():
-			return ctx.Err()
+			return false, ctx.Err()
+		case result, ok := <-statusChan:
+			// The status channel being closed means that we won't be receiving
+			// any more information from that synchronizer and we should
+			// probably shut it down.
+			if !ok {
+				return true, nil
```
tanderson-ld: Does this necessarily indicate it should be removed?

internal/datasystem/fdv2_datasystem.go
```diff
@@ -334,7 +334,7 @@ func (f *FDv2) consumeSynchronizerResults(
 			// any more information from that synchronizer and we should
 			// probably shut it down.
```
tanderson-ld: Is this comment still accurate?

tanderson-ld (approved):

PR #259, opened 2025-04-21
chore: Add fallback to FDv1 support

keelerm84:


internal/datasourcev2/fdv1_polling_data_source.go
```diff
@@ -0,0 +1,88 @@
...
+		}
+
+		for _, keyedItem := range item.Items {
+			if keyedItem.Item.Item == nil {
+				continue
+			}
+
+			bytes, err := json.Marshal(keyedItem.Item.Item)
+			if err != nil {
+				r.loggers.Warn("Error marshalling v1 item to JSON: %s", err)
```
tanderson-ld, line 59: If the entire message can't be converted, might it make sense to return an error for the whole message?  Partial states may undesirable.

tanderson-ld: And I think bytes may be indeterminate in this current code without a continue.

internal/datasourcev2/polling_data_source.go
```diff
@@ -118,6 +118,16 @@ func (pp *PollingProcessor) Sync() <-chan interfaces.DataSourceStatus {
 							StatusCode: hse.Code,
 							Time:       time.Now(),
 						}
+
+						if hse.Header.Get("X-LD-FD-Fallback") != "" {
```
tanderson-ld: Consider checking value in case FD includes it but false.

interfaces/data_source_status_provider.go
```diff
@@ -83,6 +83,16 @@ type DataSourceStatusProvider interface {
 	WaitFor(desiredState DataSourceState, timeout time.Duration) bool
 }
 
+// DataSynchronizerStatus represents intermediate information that a data synchronizer may provide
+// about its status.
+//
+// Synchronizers are expected to emit these status updates on the channel returned by Sync().
+type DataSynchronizerStatus struct {
```
tanderson-ld, line 90: Is this specific to synchronizers and not initializers?

internal/datasourcev2/streaming_data_source.go
```diff
@@ -362,6 +358,16 @@ func (sp *StreamProcessor) subscribe(statusChan chan<- interfaces.DataSourceStat
 				StatusCode: se.Code,
 				Time:       time.Now(),
 			}
+
+			if se.Header.Get("X-LD-FD-Fallback") != "" {
```
tanderson-ld: Same comment about checking value here.

internal/datasystem/fdv2_datasystem.go
```diff
@@ -271,6 +275,10 @@ func (f *FDv2) runSynchronizers(ctx context.Context, closeWhenReady chan struct{
 				f.primarySyncBuilder = f.secondarySyncBuilder
 				f.secondarySyncBuilder = nil
 
+				if fallbackv1 {
+					f.primarySyncBuilder = f.fdv1SyncBuilder
+				}
+
 				if f.primarySyncBuilder == nil {
```
tanderson-ld, line 282: As long as fdv1SyncBuilder can't be null this seems fine.  If fdv1SyncBuilder can be null, the secondary sync won't be reached.

interfaces/data_source_status_provider.go
```diff
@@ -83,6 +83,16 @@ type DataSourceStatusProvider interface {
...
+	State    DataSourceState
+	Error    DataSourceErrorInfo
+	Fallback bool
```
tanderson-ld: Consider a name that makes this more clearly FDv1 related since "fallback" is a term used when falling back between primary and secondary sources.

ldcomponents/data_system_configuration_builder.go
```diff
@@ -47,11 +48,18 @@ func (d *DataSystemModes) Default() *DataSystemConfigurationBuilder {
...
 	if d.endpoints.Polling != "" {
 		polling.BaseURI(d.endpoints.Polling)
+		fallback.BaseURI(d.endpoints.Polling)
 	}
-	return d.Custom().Initializers(polling.AsInitializer()).Synchronizers(streaming, polling)
+
+	return d.Custom().
+		Initializers(polling.AsInitializer()).
+		Synchronizers(streaming, polling).
+		fallback(fallback)
```
tanderson-ld: Consider making fallback function more specific to FDv1 fallback to avoid confusion with already used "fallback" term.

interfaces/data_source_status_provider.go
keelerm84: Currently yes. We do not modify the data system status based on intermediate failures in the initializers. We would consider that still initializing I do believe.

interfaces/data_source_status_provider.go
keelerm84: Renamed.

internal/datasourcev2/fdv1_polling_data_source.go
keelerm84: I have updated it to fail the entire payload.

internal/datasourcev2/polling_data_source.go
keelerm84: Done.

internal/datasourcev2/streaming_data_source.go
keelerm84: Done.

internal/datasystem/fdv2_datasystem.go
keelerm84: I'm not sure I follow the concern here. On line 271, I promote the secondary sync to the primary slot. Then, if we are supposed to fall back to v1, I overwrite the primary again with the fallback.

Later, if we are processing the secondary one already and we receive the fallback, I can just set the primary to the fallback, and nil out the secondary.

ldcomponents/data_system_configuration_builder.go
keelerm84: Done.

internal/datasystem/fdv2_datasystem.go
tanderson-ld: Ah, yes, I was mistaken.

tanderson-ld (approved):
Consider testing against the SDK Test Harness branch we added the FDv1 fallback test to.

PR #260, opened 2025-04-23
chore: Hook up flag change listeners to FDv2

keelerm84:


internal/datasystem/store.go
```diff
@@ -258,3 +308,74 @@ func (s *Store) Get(kind ldstoretypes.DataKind, key string) (ldstoretypes.ItemDe
...
+		for key := range newItems {
+			if _, found := oldItems[key]; !found {
+				allKeys = append(allKeys, key)
+			}
+		}
+		for _, key := range allKeys {
+			oldItem, haveOld := oldItems[key]
+			newItem, haveNew := newItems[key]
+			if haveOld || haveNew {
+				if !haveOld || !haveNew || oldItem.Version < newItem.Version {
```
tanderson-ld: Double check if this version check should be here.  I recall us deciding that we wouldn't have version checks in model processing in FDv2.  There is a story in the JS FDv2 epic (SDK-1046) for verifying the version check is required / not required.  I haven't gotten to that story yet, but plan to.

internal/datasystem/store.go
```diff
@@ -161,8 +175,29 @@ func (s *Store) SetBasis(events []fdv2proto.Change, selector fdv2proto.Selector,
...
+			}
+		}
+	}
+
 	s.memoryStore.SetBasis(collections)
 
+	s.updateDependencyTrackerFromFullDataSet(collections)
+	if oldData != nil {
+		s.sendChangeEvents(s.computeChangedItemsForFullDataSet(oldData, fullDataSetToMap(collections)))
+	}
```
tanderson-ld, line 207: This if condition reads a little weird.  I wouldn't expect sending change events to require there be `oldData`.  I see that `oldData` is set when `s.flagChangeEvent.HasListeners()` is true up above.  Perhaps `oldData`'s name could be improved or the condition should refer to `s.flagChangeEvent.HasListeners()`?

keelerm84: Yeah, it is a bit weird. The problem with using the has listener check is that it might have changed between those two checks. So we can't do that.

I'm fine with improving the name if you have a suggestion. We can only calculate the difference if we have a snapshot of the old data. Maybe I can just add more comments explaining why we are doing it this way?

internal/datasystem/store.go
keelerm84: That's a good point. It shouldn't have to care about that any more. So I'll remove it.

internal/datasystem/store.go
keelerm84: Took a shot at updating it. Let me know.

tanderson-ld (approved):

PR #261, opened 2025-04-23
chore: Add waitFor implementation and expand tests

keelerm84:


internal/datasystem/fdv2_datasystem.go
```diff
@@ -488,9 +495,21 @@ func (d *dataStatusProvider) RemoveStatusListener(listener <-chan interfaces.Dat
...
-	// TODO(SDK-930): Implement dataStatusProvider for this data system.
-	panic("implement me")
+	ch := d.AddStatusListener()
+	defer d.RemoveStatusListener(ch)
+
+	timer := time.NewTimer(timeout)
+
+	for {
+		select {
+		case status := <-ch:
```
tanderson-ld, line 512: Does this complete immediately if the status is A and then waitFor(A, ...) is called?

tanderson-ld (approved):

internal/datasystem/fdv2_datasystem.go
keelerm84: Fixed.

tanderson-ld (approved):

PR #268, opened 2025-05-15
chore: Implement snapshot method on data store

keelerm84:


internal/datasystem/store.go
```diff
@@ -300,6 +301,32 @@ func (s *Store) getActive() subsystems.ReadOnlyStore {
 	return s.active
 }
 
+//nolint:revive // Implementation for ReadOnlyDataStore.
+func (s *Store) Snapshot() (
```
keelerm84, line 305: When the relay tries to grab information from the store to determine what it should play back, it needs to get all the store information along with the selector in consistent fashion. Doing so increases data consistency but also will make some implementation details easier in the relay which you will see in the related PR [here](https://github.com/launchdarkly/ld-relay/pull/501).

subsystems/read_only_store.go
```diff
@@ -30,3 +32,17 @@ type ReadOnlyStore interface {
 	// feature flag evaluations.
 	IsInitialized() bool
 }
+
+// ReadOnlyDataStore represents a read-only data store that can be used to
+// retrieve any of the SDK's supported DataKinds.
+//
+// This field is not stable, and not subject to any backwards compatibility guarantees or semantic versioning.
+// It is not suitable for production usage. Do not use it. You have been warned.
+type ReadOnlyDataStore interface {
```
keelerm84, line 41: This name is not ideal. I realize this. I would welcome suggestions but also I'm making it clear you shouldn't be doing anything with this unless you are us.

tanderson-ld: I'd go with a part of the name that describes the snapshotable-ness / transactional-ness of the read.

tanderson-ld (approved):

PR #272, opened 2025-05-30
chore: Create v2 TestDataSource that works as synchronizer / initializer

keelerm84:


testhelpers/ldtestdatav2/test_data_source_flag.go
```diff
@@ -0,0 +1,504 @@
+package ldtestdatav2
```
keelerm84, line 1: This file is a copy and paste from the v1 package. You can skip this review.

tanderson-ld (approved):

PR #289, opened 2025-06-27
feat: Add Track Series Hook support

abarker-launchdarkly:
**Requirements**

Add Track Series Hook support

### DIFF
--- a/ldhooks/hooks.go
+++ b/ldhooks/hooks.go
@@ -97,7 +97,7 @@ func (h Unimplemented) AfterEvaluation(
 	return data, nil
 }
 
-// AfterEvaluation is a default implementation of the AfterTrack stage.
+// AfterTrack is a default implementation of the AfterTrack stage.
 func (h Unimplemented) AfterTrack(
 	_ context.Context,
 	_ TrackSeriesContext,
--- a/ldhooks/hooks.go
+++ b/ldhooks/hooks.go
@@ -21,6 +21,7 @@ import (
 type Hook interface {
 	Metadata() Metadata
 	EvaluationSeries
+	TrackSeries
 }
 
 // The EvaluationSeries is composed of stages, methods that are called during the evaluation of flags.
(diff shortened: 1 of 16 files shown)


ldhooks/hooks.go
```diff
@@ -50,9 +51,19 @@ type EvaluationSeries interface {
 	) (EvaluationSeriesData, error)
 }
 
+// The TrackSeries is composed of stages, methods that are called during the tracking of events.
+type TrackSeries interface {
+	// AfterTrack is called during the execution of a track Method.
+	AfterTrack(
+		ctx context.Context,
```
abarker-launchdarkly, line 58: For reviewers: I kept the Go context to make this consistent with the evaluation series, but I'm wondering if it's really needed.  There are variation methods that take in Go context but no track methods do this.

ldhooks/track_series_context.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"github.com/launchdarkly/go-sdk-common/v3/ldcontext"
+	"github.com/launchdarkly/go-sdk-common/v3/ldvalue"
+)
+
+// TrackSeriesContext contains contextual information for the execution of stages in the evaluation series.
+type TrackSeriesContext struct {
+	context     ldcontext.Context
+	key         string
+	metricValue *float64
```
abarker-launchdarkly, line 12: For reviewers: I don't really like this, but the contract tests expect metricValue to be optional. I considered keeping this similar to CustomEventData with "hasMetric bool" and "metricValue float64" and then just converting it for the JSON serialization in the contract tests, but that also doesn't feel great.

ldhooks/hooks.go
kinyoklion: Ideally I think we should probably add a track method to allow passing in a go context. Without the go context the hook methods won't really be able to serve telemetry functions as important information is propagated in the go context.

abarker-launchdarkly: Does it make sense to do that as part of this PR?

kinyoklion: I think it probably makes sense to either do it as part of this PR, or do it at least in the same release.

abarker-launchdarkly: @kinyoklion Ctx versions of Track methods have been added

kinyoklion (approved):

PR #291, opened 2025-07-10
feat: add environment ID support

abarker-launchdarkly:
**Requirements**

Added support for environment ID.
This also bumps eventsource to v1.10.0 and go-test-helpers to v3.1.0.


internal/datasourcev2/streaming_data_source.go
```diff
@@ -73,6 +74,7 @@ type StreamProcessor struct {
 	connectionAttemptLock      sync.Mutex
 	isClosed                   atomic.Bool
 	halt                       chan struct{}
+	environmentID              atomic.Pointer[ldvalue.OptionalString]
```
keelerm84: I don't think this needs to be protected by an atomic pointer. It seems like the value is only ever read or modified within a single go routine and is thus safe from concurrent modification.

abarker-launchdarkly: This is what prompted me to make the change:
https://github.com/launchdarkly/go-server-sdk/actions/runs/16204532679/job/45751466274

keelerm84: Ah, well then ignore me!

keelerm84 (approved):

keelerm84 (approved):

PR #292, opened 2025-07-15
feat: Support latest feature flagging semantic conventions.

kinyoklion:
Updates the tracing hook to use the latest semantic conventions.

This drops using the semver components provided as part of open telemetry because that is connected to the otel SDK version being used. Updating it required updating the other otel components.

I would prefer to decouple what is effectively a few strings from the versions of the OTEL dependency being used.


kinyoklion:
This PR does NOT add `feature_flag.set.id` support as the environment ID support is currently in PR. Once that is through PR, then the hook should be updated again to add it.

keelerm84 (approved):

PR #295, opened 2025-07-17
feat: add environment ID support to ldotel

abarker-launchdarkly:
**Requirements**

Adds environment ID support to ldotel.
This also bumps minimum go version to 1.23.0 and go-server-sdk to 7.13.0.

BEGIN_COMMIT_OVERRIDE
feat: add environment ID support to ldotel
fix: bump go minimum to 1.23
END_COMMIT_OVERRIDE


ldotel/tracing_hook_test.go
```diff
@@ -131,7 +154,7 @@ func TestMultipleSpanEvents(t *testing.T) {
 	stringFlagEventAttributes := attribute.NewSet(flagEventString.Attributes...)
 	stringAttributeFlagKey, _ := (&stringFlagEventAttributes).Value("feature_flag.key")
 	assert.Equal(t, flagKey, stringAttributeFlagKey.AsString())
-	stringAttributeProviderName, _ := (&boolFlagEventAttributes).Value("feature_flag.provider.name")
+	stringAttributeProviderName, _ := (&stringFlagEventAttributes).Value("feature_flag.provider.name")
```
abarker-launchdarkly, line 157: For reviewers: this change is unrelated but I fixed what appears to be a mistake in this test.

ldotel/go.mod
```diff
@@ -1,10 +1,12 @@
 module github.com/launchdarkly/go-server-sdk/ldotel
 
-go 1.20
+go 1.23.0
```
keelerm84, line 3: Make sure you add a `BEGIN_COMMIT_OVERRIDE` block and note the fix: Bump go minimum to 1.23.

ldotel/tracing_hook_test.go
keelerm84: Were the tests failing but we just missed it? If nothing was failing, then seems like this test could use some work.

ldotel/tracing_hook.go
```diff
@@ -149,6 +164,12 @@ func (h TracingHook) AfterEvaluation(ctx context.Context, seriesContext ldhooks.
 		attribs = append(attribs, featureFlagResultVariationIndex(detail.VariationIndex.IntValue()))
 	}
 
+	if h.environmentID.IsDefined() {
+		attribs = append(attribs, featureFlagSetID(h.environmentID.StringValue()))
+	} else if seriesContext.EnvironmentID().IsDefined() {
+		attribs = append(attribs, featureFlagSetID(seriesContext.EnvironmentID().StringValue()))
+	}
```
keelerm84: Alternative way of handling this with fewer function calls.

```suggestion
	if id, ok := h.environmentID.Get(); ok {
		attribs = append(attribs, featureFlagSetID(id))
	} else if id, ok := seriesContext.EnvironmentID().Get(); ok {
		attribs = append(attribs, featureFlagSetID(id))
	}
```

keelerm84 (approved):

ldotel/tracing_hook_test.go
abarker-launchdarkly: Nothing was failing, but if you look at the whole test there is a group of assertions with `boolFlagEventAttributes` and then another group of assertions with `stringFlagEventAttributes` and it looks like the bool var was inadvertently used instead of the string var here.  It just happened to pass because provider name is always "LaunchDarkly".

PR #297, opened 2025-07-22
fix(experimental): Add experimental LDScopedClient for incrementally building and propagating contexts

aengelberg:
You can now create a "scoped client", a wrapper of the client that will use the specified context for all operations, including feature flag evaluations and event tracking. This facilitates propagating a context down to all logic that is related to a scoped piece of logic, like a web request.

An `LDScopedClient` implements most of the same methods as `LDClient`, but without the `context` argument.

A scoped client is mutable, to facilitate incrementally building up a multi-context representing the current scope. You can add new contexts (as long as they aren't replacing an existing kind) with `LDScopedClient.AddContext`, and all contexts added so far will be combined into a multi-context whenever the scoped client is used. `LDScopedClient.OverwriteContextByKind` is offered as an escape hatch if you need to update existing data.

### DIFF
--- a/ldclient_scoped.go
+++ b/ldclient_scoped.go
@@ -10,7 +10,6 @@ import (
 	"github.com/launchdarkly/go-sdk-common/v3/ldvalue"
 	ldevents "github.com/launchdarkly/go-sdk-events/v3"
 	"github.com/launchdarkly/go-server-sdk/v7/interfaces"
-	"github.com/launchdarkly/go-server-sdk/v7/interfaces/flagstate"
 )
 
 // TODO:
@@ -237,12 +236,6 @@ func (c *LDScopedClient) MigrationVariationCtx(
 	defaultStage ldmigration.Stage,
 ) (ldmigration.Stage, interfaces.LDMigrationOpTracker, error) {
 	return c.client.MigrationVariationCtx(ctx, key, c.CurrentContext(), defaultStage)
-}
-
-func (c *LDScopedClient) AllFlagsState(
-	options ...flagstate.Option,
-) flagstate.AllFlags {
-	return c.client.AllFlagsState(c.CurrentContext(), options...)
 }
 
 func (c *LDScopedClient) Identify() error {
--- a/ldclient_scoped.go
+++ b/ldclient_scoped.go
@@ -131,17 +131,18 @@ func (c *LDScopedClient) overwriteIndividualContextByKind(context ldcontext.Cont
 // OverwriteContextByKind overwrites an existing context in the scoped client, affecting all future
 // operations on it, like flag evaluations and event tracking.
 //
-// You may pass a multi-context, or multiple individual contexts, to OverwriteContextByKind.
-// Any existing contexts with the same kind as any of the new contexts will be overwritten.
-//
-// If the scoped client had multiple contexts, only the context with the same kind as the new context
-// will be overwritten. Any other existing contexts will remain unchanged:
+// If the scoped client already has multiple contexts, only the individual
+// context with the same kind as the new context will be overwritten. Any other
+// existing contexts will remain unchanged:
 //
 //	company := ldcontext.NewWithKind("company", "acme")
 //	scopedClient := ld.NewScopedClient(client, company)
 //	scopedClient.AddContext(ldcontext.New("user-key"))
 //	scopedClient.OverwriteContextByKind(ldcontext.NewWithKind("company", "monsoon"))
 //	scopedClient.CurrentContext() // returns a multi-context with "monsoon" and "user-key"
+//
+// Passing a multi-context to OverwriteContextByKind is equivalent to passing in
+// all of its individual contexts separately.
 func (c *LDScopedClient) OverwriteContextByKind(contexts ...ldcontext.Context) {
 	c.Lock()
 	defer c.Unlock()
--- a/ldclient_scoped.go
+++ b/ldclient_scoped.go
@@ -18,11 +18,11 @@ import (
 // create a new scoped client for each logical scope where the evaluation context
 // should be isolated from other scopes, like a web request.
 //
-// An LDScopedClient is created by calling [LDClient.ForContext]. This sets the
+// An LDScopedClient is created by calling [NewScopedClient]. This sets the
 // initial context to be used for all operations:
 //
 //	userContext := ldcontext.New("user-key")
-//	scopedClient := client.ForContext(userContext)
+//	scopedClient := ld.NewScopedClient(client, userContext)
 //	scopedClient.BoolVariation("flag-key", false)
 //
 // A scoped client contains one or more contexts, each with a unique kind. You
@@ -30,7 +30,7 @@ import (
 // new contexts' kinds are not already present. The "current context" is the
 // combination of all contexts added so far, as a multi-context.
 //
-//	scopedClient := client.ForContext(userContext)
+//	scopedClient := ld.NewScopedClient(client, userContext)
 //	scopedClient.CurrentContext() // returns the single "user" context
 //	scopedClient.AddContext(ldcontext.NewWithKind("company", "acme"))
 //	scopedClient.CurrentContext() // returns a multi-context with a "user" and "company" context
@@ -40,7 +40,7 @@ import (
 // available early in the request, but you also want to evaluate flags with a
 // `company` context that is only available later:
 //
-//	scopedClient := client.ForContext(ldcontext.New("user-key"))
+//	scopedClient := ld.NewScopedClient(client, ldcontext.New("user-key"))
 //	company := fetchCompanyForUser(user)
 //	scopedClient.AddContext(ldcontext.NewWithKind("company", company.Id))
 //	scopedClient.BoolVariation("enterprise-features", false)
@@ -67,22 +67,20 @@ type LDScopedClient struct {
 	rebuild bool
 }
 
-// ForContext creates a new LDScopedClient, a wrapper of LDClient that uses a
+// NewScopedClient creates a new LDScopedClient, a wrapper of LDClient that uses a
 // certain LaunchDarkly evaluation context for all operations, like evaluating
 // feature flags or sending events. The scoped client supports most of the same
 // methods as LDClient, like BoolVariation et al., but without the
 // ldcontext.Context parameter.
 //
-// You may pass a multi-context, or multiple individual contexts, to ForContext.
-// All of the contexts passed in will be combined into a multi-context when the
-// scoped client is used.
+// You may pass one or more contexts to NewScopedClient. All contexts will be
+// combined into a multi-context when the scoped client is used. Passing in one
+// multi-context is equivalent to passing in all of its individual contexts
+// separately.
 //
-// You should create a new scoped client for each logical scope where the
-// contexts are isolated from each other, like a web request.
-//
-// For more info on how to use the scoped client, see the documentation for
-// LDScopedClient.
-func (client *LDClient) ForContext(contexts ...ldcontext.Context) *LDScopedClient {
+// For more information on how to use the scoped client, read the documentation
+// for LDScopedClient.
+func NewScopedClient(client *LDClient, contexts ...ldcontext.Context) *LDScopedClient {
 	cc := &LDScopedClient{
 		client:   client,
 		contexts: make(map[ldcontext.Kind]ldcontext.Context),
@@ -140,7 +138,7 @@ func (c *LDScopedClient) overwriteIndividualContextByKind(context ldcontext.Cont
 // will be overwritten. Any other existing contexts will remain unchanged:
 //
 //	company := ldcontext.NewWithKind("company", "acme")
-//	scopedClient := client.ForContext(company)
+//	scopedClient := ld.NewScopedClient(client, company)
 //	scopedClient.AddContext(ldcontext.New("user-key"))
 //	scopedClient.OverwriteContextByKind(ldcontext.NewWithKind("company", "monsoon"))
 //	scopedClient.CurrentContext() // returns a multi-context with "monsoon" and "user-key"
--- a/ldclient_scoped.go
+++ b/ldclient_scoped.go
@@ -18,6 +18,10 @@ import (
 // create a new scoped client for each logical scope where the evaluation context
 // should be isolated from other scopes, like a web request.
 //
+// This type is not stable, and not subject to any backwards compatibility
+// guarantees or semantic versioning. It is not suitable for production usage. Do
+// not use it. You have been warned.
+//
 // An LDScopedClient is created by calling [NewScopedClient]. This sets the
 // initial context to be used for all operations:
 //
@@ -80,6 +84,10 @@ type LDScopedClient struct {
 //
 // For more information on how to use the scoped client, read the documentation
 // for LDScopedClient.
+//
+// This function is not stable, and not subject to any backwards compatibility
+// guarantees or semantic versioning. It is not suitable for production usage. Do
+// not use it. You have been warned.
 func NewScopedClient(client *LDClient, contexts ...ldcontext.Context) *LDScopedClient {
 	cc := &LDScopedClient{
 		client:   client,
--- a/ldclient_scoped.go
+++ b/ldclient_scoped.go
@@ -514,14 +514,6 @@ func (c *LDScopedClient) MigrationVariationCtx(
 	return c.client.MigrationVariationCtx(ctx, key, c.CurrentContext(), defaultStage)
 }
 
-// Identify sends an identify event for the current evaluation context.
-//
-// For more information, see the Reference Guide:
-// https://docs.launchdarkly.com/sdk/features/identify#go
-func (c *LDScopedClient) Identify() error {
-	return c.client.Identify(c.CurrentContext())
-}
-
 // TrackEvent sends a custom event for the current evaluation context.
 //
 // The eventName parameter is defined by the application and will be shown in
--- a/ldclient_scoped.go
+++ b/ldclient_scoped.go
@@ -12,9 +12,9 @@ import (
 	"github.com/launchdarkly/go-server-sdk/v7/interfaces"
 )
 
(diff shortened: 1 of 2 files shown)


ldclient_scoped.go
```diff
@@ -0,0 +1,557 @@
...
+	c.context = b.Build()
+	return c.context
+}
+
+// Client returns the underlying LDClient that this scoped client wraps.
+func (c *LDScopedClient) Client() *LDClient {
+	return c.client
+}
+
+// Contextual methods: equivalent to calling the same method on the underlying client with the current context
```
aengelberg, line 193: All of the remaining methods in this file mirror some existing method on `LDClient`, but instead of taking a `context`, use the `c.CurrentContext()`.

All the doc comments are copied over mostly verbatim - I figured this would be better than just `See LDClient.MethodName for more information`, because once a customer adopts the scoped client, there will be developers in that organization that _only_ ever interact with the scoped client, and they may need quick explainers of each method if it's their first time using LD clients at all.

ldclient_scoped.go
```diff
@@ -0,0 +1,557 @@
...
+//	scopedClient.AddContext(ldcontext.NewWithKind("company", company.Id))
+//	scopedClient.BoolVariation("enterprise-features", false)
+//
+// You can also overwrite a context that was previously added, by calling
+// [LDScopedClient.OverwriteContextByKind]. This can be used when an existing
+// context needs to be updated with new data:
+//
+//	entitlements := fetchUserEntitlements(user)
+//	fullUserCtx := ldcontext.NewBuilder(user.Id).Set("entitlements", entitlements).Build()
+//	scopedClient.OverwriteContextByKind(fullUserCtx)
```
aengelberg, line 59: I'm interested in feedback on all of the documentation I added, but particularly this big block.

ldclient_scoped.go
```diff
@@ -0,0 +1,557 @@
...
+// AddContext adds additional evaluation contexts to the scoped client, affecting
+// all future operations on it, like flag evaluations and event tracking.
+//
+// All of the contexts added must have kinds that are not already present. If you
+// try to add a context with a duplicate kind, the new context will not be added
+// and a warning will be logged. If you want to replace an existing context, use
+// [LDScopedClient.OverwriteContextByKind] instead.
+//
+// The scoped client's contexts so far are combined into a multi-context whenever the
+// scoped client is used.
```
aengelberg: I'm handwaving a bit over the fact that you can pass a multi-context into `AddContext` and that will be broken down into individual contexts before trying to add them all into the contexts seen so far. I couldn't think of a way of clarifying that without distracting a bit from the actual context-collection logic and validation (which is more important)...

ldclient_scoped.go
```diff
@@ -0,0 +1,612 @@
...
+	return c.client.MigrationVariationCtx(ctx, key, c.CurrentContext(), defaultStage)
+}
+
+// Identify sends an identify event for the current evaluation context.
+//
+// For more information, see the Reference Guide:
+// https://docs.launchdarkly.com/sdk/features/identify#go
+func (c *LDScopedClient) Identify() error {
+	return c.client.Identify(c.CurrentContext())
+}
```
aengelberg: I'm honestly not sure if `Identify` needs a "scoped" variant.

In general, scoped methods are not strictly necessary even if customers want to call them, because you can always call `scopedClient.Client().SomeMethod(..., scopedClient.CurrentContext(), ...)` as an escape hatch. Because of this I didn't copy over `AllFlagsState` or `SecureModeHash`, despite those being methods that currently take a context.

ldclient_scoped.go
```diff
@@ -0,0 +1,621 @@
...
+	"github.com/launchdarkly/go-sdk-common/v3/ldvalue"
+	ldevents "github.com/launchdarkly/go-sdk-events/v3"
+	"github.com/launchdarkly/go-server-sdk/v7/interfaces"
+)
+
+// LDScopedClient is a wrapper around LDClient that lets you specify the
+// evaluation context to be used for all operations, rather than taking the
+// evaluation context as a parameter every time you call a method.
+//
+// A LDScopedClient is created by calling [LDClient.ForContext]. This sets the
```
keelerm84:
```suggestion
// An LDScopedClient is created by calling [LDClient.ForContext]. This sets the
```

ldclient_scoped.go
```diff
@@ -0,0 +1,621 @@
...
+//	scopedClient.BoolVariation("flag-key", false)
+//
+// Scoped contexts are mutable, to facilitate incrementally building up a
+// multi-context representing the current scope. You should create a new scoped
+// client for each logical scope where the contexts are different. For instance,
+// if you are using a scoped client in an HTTP request handler, you should create
+// a new scoped client for each request. You should only use the mutable features
+// of LDScopedClient to add new contexts, not to change or remove existing data.
+// A scoped client is thread-safe, so you can safely use it from multiple
+// goroutines.
```
keelerm84: It seems odd to me that we say the scoped context is mutable, but we don't recommend you change the data. I think if we're going to take that firm of a stance, we should probably provide a little more explanation on why we think that's the right approach.

ldclient_scoped.go
keelerm84: Did we settle on a way to retrieve an existing context from the scoped client? Probably worth showing adding to an existing context instead of having to fully reconstruct it.

ldclient_scoped.go
```diff
@@ -0,0 +1,621 @@
...
+// You may pass a multi-context, or multiple individual contexts, to ForContext.
+// All of the contexts passed in will be combined into a multi-context when the
+// scoped client is used.
+//
+// You should create a new scoped client for each logical scope where the
+// contexts are isolated from each other, like a web request.
+//
+// For more info on how to use the scoped client, see the documentation for
+// LDScopedClient.
+func (client *LDClient) ForContext(contexts ...ldcontext.Context) *LDScopedClient {
```
keelerm84: What do we think about making this a `New` style function instead of a method of the `LDClient` class? Adding methods across multiple files like this is something I'd like to avoid if we can.

ldclient_scoped.go
keelerm84: You could add a line like:

"Providing a mulit-context to 'AddContext' is equivalent to adding each of the single-kind contexts individually."

ldclient_scoped.go
```diff
@@ -0,0 +1,621 @@
...
+// try to add a context with a duplicate kind, the new context will not be added
+// and a warning will be logged. If you want to replace an existing context, use
+// [LDScopedClient.OverwriteContextByKind] instead.
+//
+// The scoped client's contexts so far are combined into a multi-context whenever the
+// scoped client is used.
+func (c *LDScopedClient) AddContext(contexts ...ldcontext.Context) {
+	c.Lock()
+	defer c.Unlock()
+	c.rebuild = true
```
keelerm84, line 122: It is possible all the requested contexts could be dropped (due to conflicts), so you might be prematurely setting this to true. Minor optimization I realize.

ldclient_scoped.go
```diff
@@ -0,0 +1,621 @@
...
+}
+
+func (c *LDScopedClient) overwriteIndividualContextByKind(context ldcontext.Context) {
+	c.contexts[context.Kind()] = context
+}
+
+// OverwriteContextByKind overwrites an existing context in the scoped client, affecting all future
+// operations on it, like flag evaluations and event tracking.
+//
+// You may pass a multi-context, or multiple individual contexts, to OverwriteContextByKind.
```
keelerm84: You can also pass multiple multi-context contexts right?

ldclient_scoped.go
keelerm84: I agree. Having this method here I think would only add confusion. I think it's better to remove it entirely.

keelerm84 (commented):
Looking pretty good!

ldclient_scoped.go
aengelberg: It's pretty complicated, but you could do:
```go
ldcontext.NewBuilderFromContext(scopedClient.CurrentContext().IndividualContextByKind("user")).Set("entitlements", entitlements).Build()
```
That feels okay enough to have that be the way to do it, but complicated enough I felt weird putting it in the documentation.

ldclient_scoped.go
aengelberg: What I was going for here is:
* The scoped context is mutable, so we want to loudly disclaim that so you're aware that the multi-context could be different from one flag eval to the next
* However, you usually will only be _adding_ new context kinds, not changing existing contexts, and we recommend doing that (but it's still mutation, and that's still important to be aware of)
* But updating existing data is technically something you could do, and we don't recommend doing that

ldclient_scoped.go
```diff
@@ -0,0 +1,621 @@
...
+// if you are using a scoped client in an HTTP request handler, you should create
+// a new scoped client for each request. You should only use the mutable features
+// of LDScopedClient to add new contexts, not to change or remove existing data.
+// A scoped client is thread-safe, so you can safely use it from multiple
+// goroutines.
+//
+// You may add new contexts with [LDScopedClient.AddContext]. The scoped client's
+// contexts so far are combined into a multi-context whenever the scoped client
+// is used. This is useful when more contexts become available later in the
+// lifecycle of a request. For instance, you might have a `user` context that is
```
mmrj: Optional suggestion to attempt to clarify that the scoped client only has one multi-context

```suggestion
// You may add new contexts to the scoped client's multi-context with [LDScopedClient.AddContext]. 
// This is useful when more contexts become available later in the
// lifecycle of a request. For instance, you might have a `user` context that is
```

ldclient_scoped.go
```diff
@@ -0,0 +1,621 @@
...
+// available early in the request, but you also want to evaluate flags with a
+// `company` context that is only available later:
+//
+//	company := fetchCompanyForUser(user)
+//	scopedClient.AddContext(ldcontext.NewWithKind("company", company.Id))
+//	scopedClient.BoolVariation("enterprise-features", false)
+//
+// You can also overwrite a context that was previously added, by calling
+// [LDScopedClient.OverwriteContextByKind]. This can be used when an existing
+// context needs to be updated with new data:
```
mmrj:
```suggestion
// You can also overwrite a context that you previously added, by calling
// [LDScopedClient.OverwriteContextByKind]. Use this when you need to update an existing
// context with new data:
```
optional suggestion, updates for active voice

ldclient_scoped.go
```diff
@@ -0,0 +1,621 @@
...
+	rebuild bool
+}
+
+// ForContext creates a new LDScopedClient, a wrapper of LDClient that uses a
+// certain LaunchDarkly evaluation context for all operations, like evaluating
+// feature flags or sending events. The scoped client supports most of the same
+// methods as LDClient, like BoolVariation et al., but without the
+// ldcontext.Context parameter.
+//
+// You may pass a multi-context, or multiple individual contexts, to ForContext.
```
mmrj:
```suggestion
// You may pass a single multi-context, or multiple individual contexts, to ForContext.
```
I think?

ldclient_scoped.go
```diff
@@ -0,0 +1,621 @@
...
+// All of the contexts passed in will be combined into a multi-context when the
+// scoped client is used.
+//
+// You should create a new scoped client for each logical scope where the
+// contexts are isolated from each other, like a web request.
+//
+// For more info on how to use the scoped client, see the documentation for
```
mmrj:
```suggestion
// For more information on how to use the scoped client, read the documentation for
```

ldclient_scoped.go
```diff
@@ -0,0 +1,621 @@
...
+func (c *LDScopedClient) addIndividualContext(context ldcontext.Context) {
+	if _, ok := c.contexts[context.Kind()]; ok {
+		c.client.loggers.Warnf("Tried to add a duplicate %s context to LDScopedClient", context.Kind())
+		return
+	}
+	c.contexts[context.Kind()] = context
+}
+
+// AddContext adds additional evaluation contexts to the scoped client, affecting
+// all future operations on it, like flag evaluations and event tracking.
```
mmrj:
```suggestion
// AddContext adds additional evaluation contexts to the scoped client's multi-context. This affects
// all future operations on it, like flag evaluations and event tracking.
```
optional suggestion

ldclient_scoped.go
```diff
@@ -0,0 +1,621 @@
...
+//
+// All of the contexts added must have kinds that are not already present. If you
+// try to add a context with a duplicate kind, the new context will not be added
+// and a warning will be logged. If you want to replace an existing context, use
+// [LDScopedClient.OverwriteContextByKind] instead.
+//
+// The scoped client's contexts so far are combined into a multi-context whenever the
+// scoped client is used.
```
mmrj: I'm realizing that my mental model is that the scoped client has a _single multi-context_ that it keeps adding to. However, the documentation here is emphasizing that a scoped client has _a collection of contexts_ that are only combined into a multi-context when you call some other method on the client.

Is that difference an implementation detail, or a meaningful distinction that we should continue to clarify?

ldclient_scoped.go
mmrj: +1. This may also need clarification on line 70.

ldclient_scoped.go
```diff
@@ -0,0 +1,621 @@
...
+	c.contexts[context.Kind()] = context
+}
+
+// OverwriteContextByKind overwrites an existing context in the scoped client, affecting all future
+// operations on it, like flag evaluations and event tracking.
+//
+// You may pass a multi-context, or multiple individual contexts, to OverwriteContextByKind.
+// Any existing contexts with the same kind as any of the new contexts will be overwritten.
+//
+// If the scoped client had multiple contexts, only the context with the same kind as the new context
```
mmrj:
```suggestion
// If the scoped client already has multiple contexts, only the context with the same kind as the new context
```
optional suggestion

ldclient_scoped.go
aengelberg: @keelerm84 I've updated this whole block to not provide an ambiguous warning about mutability, just clarifying what you can do.

ldclient_scoped.go
aengelberg: I implemented this suggestion, the new function is called `NewScopedClient`

ldclient_scoped.go
aengelberg: @mmrj Hmm, I think it isn't quite right to say that the scoped client contains a multi-context. It actually builds the multi-context on the fly when you need it. Also, multi-contexts are immutable, so talking about "adding contexts to the multi-context" may raise questions about whether you're actually changing the multi-context or just the client's reference to it (which it doesn't even have). However, I think you're right that we need some consistent way to refer to the "multi-context so far" that's implicitly available in a scoped client.

I've updated the comments in this file to use the term "current context" (which matches the method `CurrentContext()`) to refer the implicit inner multi-context, then use that consistently throughout when talking about adding stuff to the current context.

```
The "current context" is the
combination of all contexts added so far, as a multi-context.
```

ldclient_scoped.go
aengelberg: > Is that difference an implementation detail, or a meaningful distinction that we should continue to clarify?

@mmrj I'm still torn on this. Right now I think it's a meaningful distinction, because of the mutability. Contexts (and multi-contexts) are supposed to be immutable, so if I say the scoped client contains a multi-context, but also I'm "adding a context" to it, that could raise some questions. I would then clarify I'm not really mutating the multi-context, I'm creating a new multi-context (though not really, because this multi-context is imaginary). And I _am_ mutating the scoped client.

I just updated the LDScopedClient documentation with this paragraph:
```
// A scoped client contains one or more contexts, each with a unique kind. You
// can add additional contexts with [LDScopedClient.AddContext], as long as the
// new contexts' kinds are not already present. The "current context" is the
// combination of all contexts added so far, as a multi-context.
```
Then I use the term "current context" whenever I mean the scoped client's current multi-context.

Does that make sense, or just make things more confusing?

ldclient_scoped.go
```diff
@@ -0,0 +1,618 @@
...
+//
+// A scoped client contains one or more contexts, each with a unique kind. You
+// can add additional contexts with [LDScopedClient.AddContext], as long as the
+// new contexts' kinds are not already present. The "current context" is the
+// combination of all contexts added so far, as a multi-context.
+//
+//	scopedClient := ld.NewScopedClient(client, userContext)
+//	scopedClient.CurrentContext() // returns the single "user" context
+//	scopedClient.AddContext(ldcontext.NewWithKind("company", "acme"))
+//	scopedClient.CurrentContext() // returns a multi-context with a "user" and "company" context
```
mmrj, line 40: this example is much clearer now, thanks

ldclient_scoped.go
```diff
@@ -0,0 +1,618 @@
...
+//
+// Adding contexts is useful when more contexts become available later in the
+// lifecycle of a request. For instance, you might have a `user` context that is
+// available early in the request, but you also want to evaluate flags with a
+// `company` context that is only available later:
+//
+//	scopedClient := ld.NewScopedClient(client, ldcontext.New("user-key"))
+//	company := fetchCompanyForUser(user)
+//	scopedClient.AddContext(ldcontext.NewWithKind("company", company.Id))
+//	scopedClient.BoolVariation("enterprise-features", false)
```
mmrj, line 50:
```suggestion
//	scopedClient.BoolVariation("enterprise-features", false) // evaluates the flag using a multi-context with "user" and "company" contexts
```
optional suggestion

ldclient_scoped.go
mmrj: this makes sense to me! thanks for the updates to use "current context" more clearly

mmrj (approved):
✅ from Docs for the in-code documentation. I did not review the implementation or tests carefully.

keelerm84 (approved):

ldclient_scoped.go
aengelberg: I think I'm going to keep this the way it is - the "suboptimal" thing will only happen if you're using this function wrong.

PR #306, opened 2025-08-13
fix(experimental): Add ldmiddleware library, utilities for creating scoped clients, contexts, events automatically

aengelberg:
Add new `ldmiddleware` library which provides four HTTP middlewares that can help jumpstart a LaunchDarkly setup:

* `AddScopedClientForRequest` - creates a scoped client and adds a `ld_request`-kind context to it, an anonymous context populated with various attributes of the incoming HTTP request
* `AddScopedClientForRequestWithKeyFn` - same except you provide a custom function for determining the `key` of each request context
* `TrackTiming` - track duration metrics using whatever contexts are set so far in the scoped client
* `TrackErrorResponses` - track 4xx & 5xx error responses using whatever contexts are set so far in the scoped client

See README.md for sample usage.

Note that these functions are simple enough that consumers could implement their own versions of these if they want the behavior to be slightly different. However, I've set the initial library version as `0.1.0` and am advertising it as an an "alpha" state so we're free to keep making changes.


ember-stevens (approved):

ldmiddleware/http_middleware.go
```diff
@@ -0,0 +1,107 @@
...
+			if keyFn != nil {
+				if k, ok := keyFn(r); ok && k != "" {
+					requestKey = k
+				}
+			}
+			if requestKey == "" {
+				requestKey = uuid.New().String()
+			}
+
+			b := ldcontext.NewBuilder(requestKey).Kind("ld_request").Anonymous(true)
```
keelerm84, line 46: I wonder if these should actually be anonymous. I don't think the auto env contexts are marked as anonymous. On the other hand, if you don't define a request key func, then we probably don't want the UUIDs blowing up the UI.

Probably something we can add later if people don't want them to be anonymous every time.

ldmiddleware/README.md
```diff
@@ -0,0 +1,57 @@
+LaunchDarkly HTTP Middleware
+===============================
+
+# ⛔️⛔️⛔️⛔️
+> [!CAUTION]
+> This library is a alpha version and should not be considered ready for production use while this message is visible.
```
keelerm84:
```suggestion
> This library is an alpha version and should not be considered ready for production use while this message is visible.
```

Makefile
```diff
@@ -109,6 +109,32 @@ ldai-lint: $(LINTER_VERSION_FILE)
...
+	fi
+
+ldmiddleware-lint: $(LINTER_VERSION_FILE)
+	@if [ -f go.work ]; then \
+		echo "Linting ldmiddleware with workspace"; \
+		$(LINTER) run ./ldmiddleware; \
+	else \
+		echo "Linting ldmiddleware without workspace"; \
+		cd ldmiddleware && 	../$(LINTER) run .; \
+	fi
```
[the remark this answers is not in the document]
keelerm84: This is a good call out @aengelberg.

ldmiddleware/http_middleware.go
aengelberg: I did this because we make all our request contexts `anonymous` internally. I think it's pretty useful to avoid trying to index these keys, so anonymous feels like the right thing to do IMO, for any sort of per-request context like this. I don't know what you mean by auto env contexts (I just know about `ld_device` and `ld_application`) but I assume those aren't as high-cardinality.

However, this brings up a good point which I've been meaning to try to solve - I've started a thread here: https://launchdarkly.slack.com/archives/C4LUP9NPM/p1755642372124419 While I am pretty sure customers can do anonymous request contexts and not run into any issues, our public docs kind of imply they're not supposed to do that, so I think we should revisit the docs or rephrase what exactly our restrictions are so this is considered safe.

Part of making this an alpha version initially is I'd like to potentially iterate on this as you're suggesting.

aengelberg: Confirmed my suspicion on the [thread](https://launchdarkly.slack.com/archives/C4LUP9NPM/p1755642372124419), anonymous request contexts is the best practice we want to encourage for customers.

keelerm84 (approved):

PR #330, opened 2025-11-24
fix: align fdv2 intializer to spec

joker23:
**Requirements**

- [x] I have added test coverage for new or changed functionality

This PR will make it so that the initializers will keep cycling to the first intializer to report a valid basis + selector OR if there are no initializers left. In this case, the synchronizers will start. In this new flow, synchronizers could put the client into a valid state.

Initialization phase will no longer throw an error unless the initializers timeout. 

We may want to introduce an option for customers to enable a more strict initialization flow where the client will terminate if all initializers fail, but that would need an addendum to the specs and a new PR.

**Additional context**

Add any other context about the pull request here.

---

> [!NOTE]
> FDv2 now only marks ready when an initializer provides a selector, updates full-dataset change detection to consider version increases, makes file data source emit NoSelector, and updates/extends tests including a new e2e deferral case.
> 
> - **FDv2 core**:
>   - Initializers now mark ready only if `basis.ChangeSet.Selector().IsDefined()`; otherwise continue to synchronizers.
> - **Store**:
>   - `computeChangedItemsForFullDataSet` now flags affected items when `oldItem.Version < newItem.Version` in addition to add/remove.
> - **File data source (v2)**:
>   - Change sets are finished with `subsystems.NoSelector()`; remove selector/version usage.
> - **Tests**:
>   - Update file data source tests to expect `NoSelector`.
>   - Add `TestFDV2FileInitializerWillDeferToFirstSynchronizer` end-to-end test validating deferral to the first synchronizer.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e37a5a1e88f1bfc704c3b0ae373957221c787f14. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


internal/datasystem/store.go
```diff
@@ -399,8 +399,18 @@ func (s *Store) computeChangedItemsForFullDataSet(
...
 		}
 		for _, key := range allKeys {
-			_, haveOld := oldItems[key]
-			_, haveNew := newItems[key]
+			old, haveOld := oldItems[key]
+			new, haveNew := newItems[key]
+
+			// If the flag version is changed, then we would want to refresh the value. This could happen
+			// if the we initialize from a stale data source.
+			if haveOld && haveNew {
```
joker23: Would like to get an opinion on this change... I think we are trying to figure out if there are any diffs between 2 payloads? If so I don't think the previous logic is correct (`(haveOld || haveNew) && (!haveOld || !haveNew)`).

keelerm84: You are right that we need a version check here. Can you actually make the code match what is shown in this [older copy](https://github.com/launchdarkly/go-server-sdk/blob/v7/internal/datasource/data_source_update_sink_impl.go#L288)?

The next major breaking change will see us remove that duplication, so don't worry about that bit of it.

keelerm84 (approved):

PR #358, opened 2026-03-30
chore: pin third-party GitHub Actions to commit SHAs

pkaeding:
## Summary

Pin all third-party GitHub Actions to full-length commit SHAs to prevent supply chain attacks.

Addresses findings from the [`third-party-action-not-pinned-to-commit-sha`](https://github.com/launchdarkly/semgrep-rules/blob/main/github-actions/third-party-action-not-pinned-to-commit-sha.yml) Semgrep rule.

## Test plan

- [ ] Verify CI passes with pinned action SHAs

---

> [!NOTE]
> **Low Risk**
> Low risk: workflow-only changes that should not affect runtime behavior beyond ensuring deterministic, supply-chain-hardened action versions.
> 
> **Overview**
> Pins third-party GitHub Actions to full commit SHAs in CI workflows to improve supply-chain security and reproducibility.
> 
> Updates `check-go-versions.yml` (fetch API + PR creation steps) and `release-please.yml` (release automation) to use SHA-pinned action references with version comments.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit fb3077f42e65f46915227da02dd8bf1828c0c100. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #359, opened 2026-03-31
chore: fix version designation on pinned action

pkaeding:
## Summary

Pin all third-party GitHub Actions to full-length commit SHAs to prevent supply chain attacks.

Addresses findings from the [`third-party-action-not-pinned-to-commit-sha`](https://github.com/launchdarkly/semgrep-rules/blob/main/github-actions/third-party-action-not-pinned-to-commit-sha.yml) Semgrep rule.

## Test plan

- [ ] Verify CI passes with pinned action SHAs

### DIFF
--- a/.github/workflows/check-go-versions.yml
+++ b/.github/workflows/check-go-versions.yml
@@ -63,7 +63,7 @@ jobs:
 
       - name: Create pull request
         if: steps.update-go-versions.outcome == 'success'
-        uses: peter-evans/create-pull-request@c5a7806660adbe173f04e3e038b0ccdcd758773c # v6.0.0
+        uses: peter-evans/create-pull-request@b1ddad2c994a25fbc81a28b3ec0e368bb2021c50 # v6.0.0
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           add-paths: |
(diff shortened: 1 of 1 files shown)


.github/workflows/check-go-versions.yml
```diff
@@ -63,7 +63,7 @@ jobs:
 
       - name: Create pull request
         if: steps.update-go-versions.outcome == 'success'
-        uses: peter-evans/create-pull-request@c5a7806660adbe173f04e3e038b0ccdcd758773c # v6.0.0
+        uses: peter-evans/create-pull-request@b1ddad2c994a25fbc81a28b3ec0e368bb2021c50 # v6.0.0
```
keelerm84: This PR is downgrading the version of the GHA. Should it be this instead?

```suggestion
        uses: peter-evans/create-pull-request@c5a7806660adbe173f04e3e038b0ccdcd758773c # v6.1.0
```

keelerm84 (changes requested):

.github/workflows/check-go-versions.yml
pkaeding: Thanks! I'm not sure why my script even touched this-- it was already pinned.

pkaeding: Ahh, I'm guessing the report of issues that I fed my script ran before https://github.com/launchdarkly/go-server-sdk/pull/358 was merged, and the script ran afterwards, and that is why the script tried to run on an already-pinned workflow.

I will close this PR, because it is already pinned. If you want to upgrade the version, feel free.

keelerm84 (approved):

PR #372, opened 2026-05-04
fix(ldai): Populate VariationKey and Version on Config returned by CompletionConfig

alohaninja:
## Summary

`Config.VariationKey()` and `Config.Version()` always return zero values (`""` and `1` respectively) when called on a `Config` returned from `CompletionConfig()` or `JudgeConfig()`.

pr372-ldai-config-metadata

## Problem

`ConfigBuilder.Build()` only sets `Meta.Enabled` — it does not populate `Meta.VariationKey` or `Meta.Version`. The `evaluateConfig()` method correctly parses these fields from the flag JSON, but only passes them to `newTracker()`. The `Config` struct returned to callers has no way to access the actual variation key or version.

The `Tracker` is unaffected because it receives the values directly. Code that uses `Tracker.TrackSuccess()` etc. sends correct event data. However, any code that reads `cfg.VariationKey()` to build its own event payloads will send empty values.

## Fix

- Add unexported `variationKey` and `version` fields to `ConfigBuilder` with internal setters (`withVariationKey`, `withVersion`)
- Call these setters from `evaluateConfig()` before `Build()`
- Update `Build()` to include `VariationKey` and `Version` in `Meta`
- Derive `Tracker` values from `cfg.VariationKey()` and `cfg.Version()` instead of separate local variables, keeping `Config` and `Tracker` in sync

## Tests

Added test coverage for:
- `Config.VariationKey()` and `Config.Version()` populated correctly across 5 sub-cases (both present, key only, both omitted, empty meta, explicit version 1)
- `Config` and `Tracker` emit consistent `variationKey`/`version` values in tracked events
- `JudgeConfig` also populates variation key correctly
- Default config (evaluation error path) returns expected zero values

All existing tests continue to pass.

---

> [!NOTE]
> **Low Risk**
> Low risk: change is limited to propagating LaunchDarkly `_ldMeta` fields (`variationKey`, `version`) into the returned `Config` and aligning `Tracker` construction with those values, plus added unit tests.
> 
> **Overview**
> **AI Config metadata is now preserved on returned `Config` objects.** `evaluateConfig` now sets `variationKey` and optional `version` on the `ConfigBuilder`, `Build()` writes them into `Meta`, and `newTracker` is constructed from `cfg.VariationKey()`/`cfg.Version()` to keep `Config` and emitted events consistent.
> 
> **Tests added** to cover variation/version population across meta-shape edge cases, confirm `Tracker` event payloads match the `Config` values, and verify judge/default-config behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 221371ae0fb89d85d67115b84346542f7b0f7977. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ldai/client.go
+++ b/ldai/client.go
@@ -152,6 +152,7 @@ func (c *Client) evaluateConfig(
 	}
 
 	builder := NewConfig().
+		withVariationKey(parsed.Meta.VariationKey).
 		WithModelName(parsed.Model.Name).
 		WithProviderName(parsed.Provider.Name).
 		WithEnabled(parsed.Meta.Enabled).
@@ -159,6 +160,10 @@ func (c *Client) evaluateConfig(
 		WithEvaluationMetricKey(parsed.EvaluationMetricKey).
 		WithEvaluationMetricKeys(parsed.EvaluationMetricKeys).
 		WithJudgeConfiguration(parsed.JudgeConfiguration)
+
+	if parsed.Meta.Version != nil {
+		builder.withVersion(*parsed.Meta.Version)
+	}
 
 	for k, v := range parsed.Model.Parameters {
 		builder.WithModelParam(k, v)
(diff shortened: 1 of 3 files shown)


jsonbailey:
This is currently an intentional design of the shape of the AI Configs. This would need to be a spec change if we wanted to include that information in the AI Config object. Did you have a specific use case for it?

PR #373, opened 2026-05-08
feat: Drop persistent-store cache after FDv2 in-memory store init

keelerm84:
## Summary

With FDv2 the in-memory store retains every flag and segment, so the persistent-store wrapper's `go-cache` instance becomes dead weight once `setBasis` flips the active store to memory. This change drops the cache at that point to reclaim the in-memory footprint and to eliminate the `CacheForever` leak.

- `persistentDataStoreWrapper` now holds its cache behind `atomic.Pointer[cache.Cache]`. Every method does a single `Load` at the top of the call and works with that snapshot, so concurrent reads never block on a mutex and the previous status-poller-vs-Store.mu race is closed by atomic semantics. `DropCache` is a one-shot `Swap(nil)` followed by `Flush` on the returned cache; safe to call multiple times.
- The FDv2 store coordinator captures any `DropCache`-capable persistent store via an `interface{ DropCache() }` assertion and invokes the drop in `setBasis`, between `s.active = s.memoryStore` and the persistent `Init()` call. Order matters: dropping first means the wrapper's `Init` skips the cache-rebuild loop on the cache we are about to discard. Custom `subsystems.DataStore` implementations that don't satisfy the interface continue to work unchanged.
- Godoc on `CacheTime` / `CacheSeconds` / `CacheForever` / `NoCaching` now notes that under FDv2 the cache only governs the bootstrap window and may be deprecated in a future major version. No public API change.
- New tests cover the wrapper's drop semantics (idempotency, write-through after drop) and the coordinator's invocation (drop on full payload, no drop on deltas, no-op for non-cache-dropping stores). All `-race` clean.

---

> [!NOTE]
> **Medium Risk**
> Changes core datastore read/write and bootstrap ordering around caching and persistence; behavior is well-tested but affects flag data paths during startup and outage recovery.
> 
> **Overview**
> Under **FDv2**, once the in-memory store is initialized, the persistent data store wrapper no longer keeps a long-lived `go-cache`. The wrapper stores the cache in an **`atomic.Pointer`** so **`DropCache()`** can **`Swap(nil)`** and flush without blocking concurrent readers; each operation uses one loaded snapshot for the whole call.
> 
> The **datasystem `Store`** detects persistent implementations with **`DropCache()`** and calls it from **`setBasis`** (full payload only)—after switching reads to memory and **before** persistent **`Init`**, so the wrapper does not rebuild a cache that is about to be discarded. **`Close`** on the wrapper also drops the cache.
> 
> **Godoc** on **`CacheTime`**, **`CacheForever`**, and related builder options notes that under FDv2 those settings only affect the bootstrap window. Tests cover drop idempotency, post-drop write-through, coordinator invocation on full vs delta payloads, and stores without **`DropCache`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ddcd663cf50a4fb34a1012fe1e3b55563e77a534. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/datastore/persistent_data_store_wrapper.go
+++ b/internal/datastore/persistent_data_store_wrapper.go
@@ -76,7 +76,7 @@ func (w *persistentDataStoreWrapper) Init(allData []st.Collection) error {
 	if c != nil {
 		c.Flush()
 	}
-	if err != nil && !(c != nil && w.cacheTTL < 0) {
+	if err != nil && (c == nil || w.cacheTTL >= 0) {
 		// If the underlying store failed to do the update, and we've got an expiring cache, then:
 		// 1) We shouldn't update the cache, and
 		// 2) We shouldn't be considered initialized.
@@ -191,7 +191,7 @@ func (w *persistentDataStoreWrapper) Upsert(
 	// but then suddenly fall back to old data when the cache expires. The exception is infinite-TTL
 	// mode, where we keep the cache in sync regardless so it can repopulate the store after a recovered
 	// outage.
-	if err != nil && !(c != nil && infinite) {
+	if err != nil && (c == nil || !infinite) {
 		return updated, err
 	}
 	if c == nil {
@@ -318,11 +318,6 @@ func (w *persistentDataStoreWrapper) pollAvailabilityAfterOutage() bool {
 		// in infinite cache mode, we set it even if the database update failed.
 	}
 	return true
-}
-
-// hasInfiniteCache returns true if the wrapper currently holds a cache configured with infinite TTL.
-func (w *persistentDataStoreWrapper) hasInfiniteCache() bool {
-	return w.cache.Load() != nil && w.cacheTTL < 0
 }
 
 func dataStoreCacheKey(kind st.DataKind, key string) string {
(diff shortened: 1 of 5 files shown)


internal/datastore/persistent_data_store_wrapper.go
```diff
@@ -256,46 +271,55 @@ func (w *persistentDataStoreWrapper) IsStatusMonitoringEnabled() bool {
...
 	return w.core.Close()
 }
 
+// DropCache flushes and releases the in-memory cache. Called once the FDv2
+// in-memory store has been initialized and is the source of truth. Safe to
+// call multiple times.
+func (w *persistentDataStoreWrapper) DropCache() {
+	if c := w.cache.Swap(nil); c != nil {
+		c.Flush()
+		w.loggers.Debug("Persistent store cache dropped; in-memory store is now active")
```
tanderson-ld: This log "knows too much".

```suggestion
		w.loggers.Debug("Persistent store cache dropped.")
```

tanderson-ld (approved):

PR #405, opened 2026-07-07
refactor(internal): extract shared file data loading into internal/filedata

kinyoklion:
`ldfiledata` and `ldfiledatav2` each carried their own copies of the file reading, JSON/YAML detection, flag-value expansion, absolute-path resolution, and multi-file merge logic. This extracts one implementation into a new `internal/filedata` package and ports both data sources onto it. A file-based override source (OVERRIDE spec) is about to become a third consumer of the same logic.

The shared `Merge` produces a consumer-neutral result (ordered flag and segment item lists); each data source adapts it to its own output in a few lines — v1 into store collections for `Init`, v2 into a full-transfer change set. Merge order is now deterministic for both (v1 previously emitted map-iteration order, which nothing depended on). The duplicate-key error now says `flag`/`segment` in both packages; the two previously disagreed (`features` vs `flag`).

Two ldfiledatav2 fixes ride along:
- `closeReloaderCh` was declared and passed to the reloader factory but never assigned, so `Close()` never stopped the file watcher goroutine, and the reload log line was unreachable. It is now created before the reloader starts, matching ldfiledata.
- `Fetch` spun up a goroutine and two broadcaster listeners that were never removed (a leak on every call), with a `select`/`default` that could silently return an empty no-changes basis instead of the file data. It now loads the files synchronously and returns the change set or an error directly. Read/parse failures still report error kind `UNKNOWN` and merge failures `INVALID_DATA`, as before.

No public API changes. All existing ldfiledata/ldfiledatav2/ldfilewatch tests pass unchanged; the shared package has its own unit tests.

SDK-2654

---

> [!NOTE]
> **Medium Risk**
> Touches core file-based flag loading for v1 and v2 data sources and changes Fetch/reload behavior in v2, though behavior is intended to match prior semantics with bug fixes.
> 
> **Overview**
> Duplicates file read/parse/merge logic from **`ldfiledata`** and **`ldfiledatav2`** move into new **`internal/filedata`** (`ReadFile`, `AbsFilePaths`, `Merge`, flag-value expansion). Each data source maps the neutral **`MergeResult`** into its own store init or full-transfer change set. **Merge** now keeps **first-seen document order** (v1 previously used map iteration); duplicate-key errors consistently say **`flag`** / **`segment`**.
> 
> **`ldfiledatav2`** also fixes: **`closeReloaderCh`** is created when the reloader starts so **`Close`** stops the watcher; **`Fetch`** loads files synchronously via **`load()`** instead of leaking broadcaster listeners and sometimes returning an empty basis.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5433c04a6dc6bfb6df5a87b252a372b2e09b018a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ldfiledata/file_data_source_impl.go
+++ b/ldfiledata/file_data_source_impl.go
@@ -22,7 +22,12 @@ type fileDataSource struct {
 	readyCh               chan<- struct{}
 	readyOnce             sync.Once
 	closeOnce             sync.Once
-	closeReloaderCh       chan struct{}
+	// closeReloaderCh is created up front rather than when the reloader starts, so that
+	// Close never races with Start assigning it.
+	closeReloaderCh chan struct{}
+	// reloaderStarted means reload calls may now come from the reloader, which is worth a
+	// log line; the initial load is not.
+	reloaderStarted bool
 }
 
 func newFileDataSourceImpl(
@@ -44,6 +49,7 @@ func newFileDataSourceImpl(
 		duplicateKeysHandling: duplicateKeysHandling,
 		reloaderFactory:       reloaderFactory,
 		loggers:               context.GetLogging().Loggers,
+		closeReloaderCh:       make(chan struct{}),
 	}
 	fs.loggers.SetPrefix("FileDataSource:")
 	return fs, nil
@@ -66,7 +72,7 @@ func (fs *fileDataSource) Start(closeWhenReady chan<- struct{}) {
 
 	// If there is a reloader, and if we haven't yet successfully loaded data, then the
 	// readiness signal will happen the first time we do get valid data (in reload).
-	fs.closeReloaderCh = make(chan struct{})
+	fs.reloaderStarted = true
 	err := fs.reloaderFactory(fs.absFilePaths, fs.loggers, fs.reload, fs.closeReloaderCh)
 	if err != nil {
 		fs.loggers.Errorf("Unable to start reloader: %s\n", err)
@@ -77,7 +83,7 @@ func (fs *fileDataSource) Start(closeWhenReady chan<- struct{}) {
 // and update the feature flag state. If any file cannot be loaded or parsed, the flag state will not
 // be modified.
 func (fs *fileDataSource) reload() {
-	if fs.closeReloaderCh != nil {
+	if fs.reloaderStarted {
 		fs.loggers.Info("Reloading flag data after detecting a change")
 	}
 	docs := make([]filedata.Document, 0)
@@ -131,9 +137,7 @@ func (fs *fileDataSource) signalStartComplete(succeeded bool) {
 // Close is called automatically when the client is closed.
 func (fs *fileDataSource) Close() (err error) {
 	fs.closeOnce.Do(func() {
-		if fs.closeReloaderCh != nil {
-			close(fs.closeReloaderCh)
-		}
+		close(fs.closeReloaderCh)
 	})
 	return nil
 }
--- a/ldfiledata/file_data_source_impl.go
+++ b/ldfiledata/file_data_source_impl.go
@@ -1,25 +1,15 @@
 package ldfiledata
 
 import (
-	"encoding/json"
-	"fmt"
-	"os"
-	"path/filepath"
-	"strings"
 	"sync"
 	"time"
-	"unicode"
 
 	"github.com/launchdarkly/go-sdk-common/v3/ldlog"
-	"github.com/launchdarkly/go-sdk-common/v3/ldvalue"
-	"github.com/launchdarkly/go-server-sdk-evaluation/v3/ldbuilders"
-	"github.com/launchdarkly/go-server-sdk-evaluation/v3/ldmodel"
 	"github.com/launchdarkly/go-server-sdk/v7/interfaces"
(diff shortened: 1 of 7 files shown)


kinyoklion:
Addressed the multi-agent review findings in 1c754e7:

- **Finding 1 (regression coverage)**: added `TestCloseStopsReloader` to both ldfiledata and ldfiledatav2 — a reloader factory captures the close channel and the test asserts `Close()` closes it. Reverting either lifecycle fix now fails a test.
- **Finding 2 (`closeReloaderCh` race)**: the channel is now created in the constructor rather than lazily in `Sync`/`Start`, so `Close` never races an assignment; an explicit flag gates the change-driven reload log instead of the channel's nil-ness. Applied to ldfiledata as well, which had the same (pre-existing) shape.
- **Finding 3 (`version` race)**: the counter is `atomic.Int64` now that both `Fetch` and the reload path go through the same load code.
- **Finding 4 (ordering doc)**: `MergeResult`'s doc now states the guarantee precisely — deterministic at document granularity only, within-document order unspecified, consumers must key by `Key`.

Findings 5–7 are informational/pre-existing and intentionally left alone. Fixes propagated through the stacked branches (#406, #407) and the `rlamb/overrides` integration branch; all file-source tests pass under `-race`, and the flag-overrides contract suite still passes end-to-end.

ldfiledatav2/file_data_source_impl.go
```diff
@@ -62,6 +58,7 @@ func newFileDataSourceImpl(
 		duplicateKeysHandling: duplicateKeysHandling,
 		reloaderFactory:       reloaderFactory,
 		loggers:               context.GetLogging().Loggers,
+		closeReloaderCh:       make(chan struct{}),
```
kinyoklion, line 61: Fix: this channel was declared and passed to the reloader factory but never assigned, so the factory always received nil and `Close()` never stopped the watcher goroutine. Creating it up front (rather than lazily in `Sync`) also means `Close` cannot race an assignment.

ldfiledatav2/file_data_source_impl.go
```diff
@@ -136,233 +134,99 @@ func (fs *fileDataSource) Sync(ds subsystems.DataSelector) <-chan subsystems.Dat
...
+		Persist:   false,
+	}, false, nil
 }
 
 // Reload tells the data source to immediately attempt to reread all of the configured source files
 // and update the feature flag state. If any file cannot be loaded or parsed, the flag state will not
 // be modified.
 func (fs *fileDataSource) reload() {
-	if fs.closeReloaderCh != nil {
+	if fs.reloaderStarted {
```
kinyoklion, line 151: Fix: this log was previously gated on `closeReloaderCh != nil`, which was never true because the channel was never assigned — the line was unreachable. Now gated on an explicit flag so the initial load still does not log it.

ldfiledatav2/file_data_source_impl.go
```diff
@@ -136,233 +134,99 @@ func (fs *fileDataSource) Sync(ds subsystems.DataSelector) <-chan subsystems.Dat
 }
 
 func (fs *fileDataSource) Fetch(ds subsystems.DataSelector, ctx context.Context) (*subsystems.Basis, bool, error) {
```
kinyoklion, line 136: Fix: `Fetch` previously spawned a goroutine and added two broadcaster listeners it never removed (a leak on every call), and its `select`/`default` could return an empty no-changes basis instead of the file data, marking an initializer initialized with no data. It now reads and merges synchronously and returns the change set or an error.

ldfiledatav2/file_data_source_impl.go
```diff
@@ -2,42 +2,38 @@ package ldfiledatav2
...
 
 type fileDataSource struct {
 	changeSetBroadcaster *internal.Broadcaster[subsystems.ChangeSet]
 	statusBroadcaster    *internal.Broadcaster[interfaces.DataSynchronizerStatus]
 	// NOTE: this is not really used anymore because file data sources at this
-	// moment will not report a selector.
-	version int
+	// moment will not report a selector. It is atomic because loads can happen
+	// concurrently from Fetch and from the reloader.
+	version atomic.Int64
```
kinyoklion, line 25: Fix: was a plain `int` incremented from both `Fetch` (caller goroutine) and the reload path — a data race now that both share one load routine. Atomic preserves the `Payload.Target` progression.

ldfiledata/file_data_source_impl.go
```diff
@@ -54,6 +49,7 @@ func newFileDataSourceImpl(
 		duplicateKeysHandling: duplicateKeysHandling,
 		reloaderFactory:       reloaderFactory,
 		loggers:               context.GetLogging().Loggers,
+		closeReloaderCh:       make(chan struct{}),
```
kinyoklion, line 52: Hardening, same shape as the v2 fix: previously assigned in `Start`, which `Close` could race. No observed bug here (Start/Close are sequential in practice), but the channel is now created up front for the same reason.

kinyoklion (commented):
Inline notes on the behavior fixes in this PR (vs. the original code).

keelerm84 (approved):

PR #410, opened 2026-07-07
feat: Serve flag overrides while uninitialized and mark override-affected evaluations in events

kinyoklion:
Completes the client-side behavior of the override system:

- **Not-initialized short-circuit**: when the client has no data from LaunchDarkly and an override source is configured, the per-key evaluation gate reads the flag through the overlay first. A hit is served, including prerequisite and segment resolution against the merged view. A miss returns the not-ready default exactly as before, and removing an override restores the short-circuit on the next evaluation. There is no separate existence check, so an evaluation cannot straddle a check and a read across a layer replacement. Whether an override source is configured is recorded once at construction; a client without one never reads the store before initialization, and the initialized hot path is untouched.
- **AllFlagsState**: when uninitialized with an override source configured, the flags are read from the merged view first. A non-empty result builds a valid state containing only the overridden flags; an empty result returns the invalid-empty state as before. The warning that only override entries were returned is logged once per client, like the cached-data warning. This mirrors the existing behavior of serving known-incomplete cached data as valid, and keeps overrides useful for bootstrap endpoints during an outage. With no overrides, the invalid-empty behavior is unchanged. (The spec doesn't prescribe this case; flagged as a decision.)
- **Events**: the evaluation's `overrideAffected` marking, taken from the evaluator's `Result.OverrideAffected` scalar, rides into `FlagEventProperties.OverrideAffected` for both top-level and prerequisite evaluations. The marking is direct or transitive: it is set when the evaluated flag, a prerequisite at any depth, or a segment read during the evaluation came from the override store, and a prerequisite record is marked by what its own subtree read. Event generation keys on that scalar alone, not on the flag definition's marker and not on the reason: a marked evaluation produces no individual feature or debug event, and the summarizer counts it under a separate counter that carries the `overrideAffected` marker (go-sdk-events).
- **Wrong-type results**: when a typed variation method replaces the result with the caller's default and a `WRONG_TYPE` error reason, the new reason keeps the evaluation's `overrideAffected` marking and the event record stays marked, matching the spec's rule that a failed evaluation is still marked when it read an override.

Client-level tests cover the gate matrix (overridden vs. not, with removal), the AllFlagsState three-way branch, the marked evaluation reason end-to-end, flag-change notifications firing when overrides change on a running client, and the marking policy: a LaunchDarkly flag whose prerequisite chain reaches an overridden flag two levels down produces marked records for the flag and the intermediate prerequisite and an unmarked record for the unaffected sibling prerequisite; with event tracking on for every flag, only the sibling appears as an individual event in the output payload while the marked flags appear only as summary counters carrying the marker; and the wrong-type path keeps the marking on both the reason and the record.

go.mod points at pseudo-versions of go-server-sdk-evaluation, go-sdk-events, and go-sdk-common that carry the `overrideAffected` API; those get pinned to real releases at the end of the series.

Based on #409.

SDK-2655

---

> [!NOTE]
> **Overview**
> **Uninitialized clients with an override source** can now evaluate and bootstrap flags from the override layer instead of always failing closed. `LDClient` records `overridesConfigured` at construction (via `FDv2.OverrideSourceConfigured()`), and when LaunchDarkly data is not yet available it still reads the merged store: overridden keys evaluate normally (with `overrideAffected` on reasons), while missing keys keep `ErrClientNotInitialized`. `AllFlagsState` follows the same split—valid state with only override entries when non-empty, otherwise the previous empty/invalid behavior, with one-time warnings.
> 
> **Analytics** propagates `Result.OverrideAffected` into evaluation events (including prerequisites and `WRONG_TYPE` coercion), aligned with the bumped `go-sdk-events` dependency. Marked evaluations are excluded from individual feature events and counted in summaries with an `overrideAffected` marker.
> 
> New `ldclient_overrides_test.go` covers the uninitialized gate, `AllFlagsState`, transitive prerequisite marking, summary vs individual event output, and flag-change notifications.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 899f3bd7a86b973feb3cb8b7324ebf8432ff9fdb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ldclient.go
```diff
@@ -1385,26 +1415,32 @@ func (client *LDClient) evaluateInternal(
...
 		errKind ldreason.EvalErrorKind,
-		flag *ldmodel.FeatureFlag,
 		err error,
 	) (ldeval.Result, *ldmodel.FeatureFlag, error) {
 		detail := newEvaluationError(defaultVal, errKind)
 		if client.logEvaluationErrors {
 			client.loggers.Warn(err)
 		}
-		return ldeval.Result{Detail: detail}, flag, err
+		return ldeval.Result{Detail: detail}, nil, err
```
keelerm84: Is it not a problem that we are always returning nil here for the flag?

ldclient_overrides_test.go
```diff
@@ -0,0 +1,465 @@
...
+
+func TestNonOverriddenFlagStillShortCircuitsWhenClientIsNotInitialized(t *testing.T) {
+	source := sharedtest.NewTestOverrideSource(
+		overrideTestFlagData(singleValueFlag("overridden-flag", ldvalue.Bool(true))))
+	client := makeUninitializedClientWithOverrides(t, source, nil)
+
+	value, detail, err := client.BoolVariationDetail("other-flag", evalTestUser, false)
+	assert.Equal(t, ErrClientNotInitialized, err)
+	assert.False(t, value)
+	assert.Equal(t, ldreason.NewEvalReasonError(ldreason.EvalErrorClientNotReady), detail.Reason)
```
keelerm84: we should add `assert.False(t, detail.Reason.IsOverrideAffected())` here as well.

ldclient.go
kinyoklion: This callback is always called on  a path where the flag was not available. Things like a malformed flag are on a different code path.

kinyoklion: I am seeing why it wasn't already producing a linting error, but starting having one when an additional callsite was introduced.

ldclient_overrides_test.go
kinyoklion: Done. It was checked in the deep equality already, but this makes it more clear.

keelerm84 (approved):

PR #411, opened 2026-07-07
test: run the OVERRIDE spec test vectors against the full client stack

kinyoklion:
Adds a runner for the eight test vectors published with the OVERRIDE spec (schema 0.4.0), vendored verbatim under `testdata/override-vectors/` with a schema-version assertion so drift in the source vectors fails loudly rather than silently changing coverage.

Each vector exercises the full production stack (client gate, overlay store, evaluator, and events input) rather than a reimplementation: LaunchDarkly data is supplied by a test initializer as a full-transfer basis with a defined selector; the `initialized: false` vectors use a synchronizer that never delivers data (with no sources configured at all, the client would consider cached data available instead of applying its not-initialized handling, so an idle synchronizer is load-bearing); overrides come from the programmatic test source. Assertions follow the vectors' comparison rules: values by JSON equality, a null variation index as undefined, reasons compared only on the fields each vector specifies with the tri-state `overrideAffected` collapse (an expected reason that omits it requires the actual reason to report false or omit it), and the `summaryOverrideAffected` expectation checked against the `OverrideAffected` scalar the client hands to the event processor for that evaluation, not against the reason.

All eight vectors (precedence, uninitialized, passthrough x2, full-evaluation, merged-view, marking, error-handling) pass, including the error-handling vector, which expects a malformed override's error result to stay marked, and the marking vector, which expects a LaunchDarkly flag to be marked when one of its rules reads an overridden segment.

Based on #410.

SDK-2655

---

> [!NOTE]
> **Overview**
> Adds **OVERRIDE spec conformance tests** at the `ldclient` layer: `TestOverrideSpecVectors` loads eight vendored vectors (`testdata/override-vectors/vectors.json`, schema **0.4.0**) and fails if the file’s `schemaVersion` drifts without updating the runner.
> 
> Each vector builds a real client via `MakeCustomClient` with a custom `DataSystem` (override source plus either a **full-transfer initializer** for `initialized: true` or the existing **hanging synchronizer** for uninitialized cases), then evaluates one flag with `JSONVariationDetail`. Assertions follow the vectors’ rules: expected value and variation index (null → undefined), partial **reason** field matching with tri-state `overrideAffected` collapse, and **`summaryOverrideAffected`** against `EvaluationData.OverrideAffected` on the capturing event processor when specified.
> 
> Coverage spans precedence, serving overrides when uninitialized, LD passthrough vs `CLIENT_NOT_READY`, full override evaluation, merged segment resolution, segment-override marking, and malformed override error handling.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d8084568998b654eaf2a52f0d1e6a6b2e58c8c47. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ldclient_override_vectors_test.go
+++ b/ldclient_override_vectors_test.go
@@ -1,0 +1,220 @@
+package ldclient
+
+import (
+	"context"
+	"encoding/json"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/launchdarkly/go-sdk-common/v3/ldcontext"
+	"github.com/launchdarkly/go-sdk-common/v3/ldlogtest"
+	"github.com/launchdarkly/go-sdk-common/v3/ldvalue"
+	ldevents "github.com/launchdarkly/go-sdk-events/v3"
+	"github.com/launchdarkly/go-server-sdk-evaluation/v3/ldbuilders"
+	"github.com/launchdarkly/go-server-sdk/v7/internal/datakinds"
+	"github.com/launchdarkly/go-server-sdk/v7/internal/sharedtest"
+	"github.com/launchdarkly/go-server-sdk/v7/internal/sharedtest/mocks"
+	"github.com/launchdarkly/go-server-sdk/v7/ldcomponents"
+	"github.com/launchdarkly/go-server-sdk/v7/subsystems"
+	"github.com/launchdarkly/go-server-sdk/v7/subsystems/ldstoreimpl"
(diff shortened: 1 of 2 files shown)


keelerm84 (approved):
Overall looks good. I think the bugbot review has a point, so we should look at that.

PR #426, opened 2026-07-27
perf: Parse FDv2 payloads in a single jsonstream pass

keelerm84:
The FDv2 data sources previously scanned every item's JSON multiple times:
the polling path ran encoding/json reflection over the whole envelope
(copying each event into a RawEvent), unmarshaled each event again into
PutObject (copying the item JSON a second time), and finally parsed the
model with jreader in ChangeSet.Collections(). The streaming path did the
same minus the envelope pass. This cost lands on every client init, since
the default data system mode runs a polling initializer first.

Replace the reflection decodes with jsonstream decoders that walk each
payload once: scalars are read in place, each put-object's item is model-
decoded as soon as it is reached, and its raw JSON is captured as a
zero-copy slice of the input via the new jreader RawValue API. The decoders
are order-independent; a data property arriving before the event name (or
an object before its kind) is captured raw and decoded once the type is
known. Unknown kinds keep their raw bytes and are now explicitly validated,
preserving the previous malformed-payload semantics.

ChangeSetBuilder gains AddParsedPut so the parsed items travel with the raw
changes, and Finish pre-populates the ChangeSet's collections cache when
every recognized put is parsed, making Collections() allocation-free.
Change.Object raw bytes are preserved for consumers such as Relay, which
re-serializes them downstream verbatim.

Benchmark (2000 flags, ParsePollingPayload + Collections):
old 21.7ms / 8.12MB / 64.1k allocs -> new 10.3ms / 4.90MB / 44.1k allocs
(2.1x faster, -40% bytes, -31% allocations). One polling-payload parse
error kind changed: per-event JSON errors now surface as invalid-data
(malformedJSONError) instead of a network error, matching the envelope
error handling.

---

> [!NOTE]
> **Medium Risk**
> Touches core FDv2 init/sync parsing on every SDK start; behavior is heavily tested for parity, but polling now classifies some per-event JSON failures as invalid-data (`malformedJSONError`) instead of network errors, and default-build change sets may alias the response body until released.
> 
> **Overview**
> **FDv2 polling and streaming** no longer decode payloads with repeated `encoding/json` passes over the envelope, each event, and again in `ChangeSet.Collections()`. Shared `parse*` helpers walk each payload once (default build: `jreader` with zero-copy raw capture and eager item decode; `launchdarkly_easyjson` build keeps reflection for parity).
> 
> **`ChangeSetBuilder`** gains **`AddParsedPut`** so recognized puts carry parsed items; **`Finish`** can pre-fill the collections cache when every known put is parsed, making **`Collections()`** skip a second JSON parse while **`Change.Object`** bytes stay verbatim for Relay.
> 
> Polling wires **`parsePollingPayload`**; streaming SSE events use the same decoders. **`go-jsonstream`** is bumped for **`RawValue`/offset APIs**. Tests and benchmarks lock behavior to the old reflection path (including integer parity with `encoding/json`, deep-nesting rejection, and ~2× faster init on large payloads).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b6fe049435b88b073f07f2830d49bef6ecadb0c9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/datasourcev2/event_parsing.go
+++ b/internal/datasourcev2/event_parsing.go
@@ -1,0 +1,220 @@
+package datasourcev2
+
+import (
+	"encoding/json"
+	"errors"
+
+	"github.com/launchdarkly/go-jsonstream/v3/jreader"
+	"github.com/launchdarkly/go-server-sdk/v7/subsystems"
+	"github.com/launchdarkly/go-server-sdk/v7/subsystems/ldstoretypes"
+)
+
+// This file contains the decoders for the FDv2 protocol events that are shared by both build
+// variants. The payload-level decoders -- the ones responsible for capturing each put-object
+// item's raw JSON alongside its parsed form -- are split by build tag:
+//
+//   - event_parsing_default.go: a single-pass jsonstream decoder using jreader.RawValue for
+//     zero-copy raw capture. Each event, and in particular each item's JSON, is scanned once.
+//   - event_parsing_easyjson.go: the previous reflection-based decode (kept for the
+//     launchdarkly_easyjson build so that no new code depends on the easyjson token reader,
+//     which is planned for removal).
+//
+// Both variants produce the same results: a parsedPutObject carrying the item's raw bytes and,
+// for recognized kinds, its parsed form, so that the ChangeSet's collections are assembled
+// without a later re-parse. The decoders are tolerant of property ordering: nothing here assumes
+// that, for example, "kind" appears before "object" in a put-object event, even though
+// LaunchDarkly services always write them in that order.
+
+// JSON property names shared by the FDv2 protocol event decoders.
+const (
+	propEvents     = "events"
+	propEvent      = "event"
+	propData       = "data"
+	propKind       = "kind"
+	propKey        = "key"
+	propVersion    = "version"
+	propObject     = "object"
+	propReason     = "reason"
+	propState      = "state"
+	propPayloads   = "payloads"
+	propID         = "id"
+	propTarget     = "target"
+	propIntentCode = "intentCode"
+	propPayloadID  = "payloadId"
+)
+
+const errNoKnownPollingEvents = "didn't receive any known protocol events in polling payload"
+
+// parsedPutObject is the result of decoding a put-object event's data.
+type parsedPutObject struct {
+	kind    subsystems.ObjectKind
+	key     string
+	version int
+	// object is the item's raw JSON. In the default build it references the input buffer passed
+	// to parsePutObjectEventData and is valid only as long as that buffer is.
+	object json.RawMessage
+	// item is the parsed representation of object, set only when hasItem is true. hasItem is
+	// false when the kind is unrecognized, in which case the raw bytes are still retained for
+	// forwards compatibility.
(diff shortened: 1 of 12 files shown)


internal/datasourcev2/event_parsing.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+	if err := r.Error(); err != nil {
+		return intent, err
+	}
+	if !gotPayload {
+		// It is a protocol error for the payload list to be missing or empty.
+		return intent, errors.New("changeset: server-intent event has no payloads")
+	}
+	return intent, nil
+}
```
aaron-zeisler: Thinking out loud: As a reader of this code, it might help navigate these payload-parsing functions if there was an example of a payload in a block comment. I understand that's included in the tests. Maybe it also exists in the call site (`polling_http_request.go` and `streaming_data_source.go`)?

internal/datasourcev2/event_parsing_default.go
```diff
@@ -0,0 +1,233 @@
...
+	p, err := readPutObject(&r, data)
+	if err == nil {
+		err = r.Error()
+	}
+	return p, err
+}
+
+// jsonWhitespace is the set of insignificant whitespace bytes RFC 8259 allows between tokens; a
+// span captured via reader offsets may include them around the value and they are trimmed off.
+const jsonWhitespace = " \t\r\n"
```
aaron-zeisler, line 209: Is there a built-in (or available 3rd-party) helper function or constant that we could use here?

internal/datasourcev2/event_parsing_easyjson.go
```diff
@@ -0,0 +1,95 @@
+//go:build launchdarkly_easyjson
+// +build launchdarkly_easyjson
```
aaron-zeisler, line 2: Question: This code is destined to be thrown away. Is it necessary because jsonstream uses easyjson 100% when the `launchdarkly_easyjson` build tag is specified? In other words, does the code in `event_parsing_default.go` work at all if the build tag is used? It seems like we would only want to use `event_parsing_default.go`, if possible.

internal/datasourcev2/event_parsing.go
aaron-zeisler: It looks like the payload structure is documented well in the sdk-specs repository: https://github.com/launchdarkly/sdk-specs/tree/main/specs/FDV2PL-payload-communication. And that will be the source of truth for what the payload should look like.

internal/datasourcev2/event_parsing_default.go
aaron-zeisler: More specifically, is it possible/safe to use `strings.TrimSpace()` or `bytes.TrimSpace()` here?

internal/datasourcev2/event_parsing_easyjson.go
aaron-zeisler: OK, I see that it's necessary as long as go-server-sdk uses a version of jsonstream w/ easyjson 😞

PR #427, opened 2026-07-28
perf(ldotel): skip building the span event when no span will record it

jstncbllr:
## Summary

`TracingHook.AfterEvaluation` builds its span-event attribute slice — including the flag value's JSON encoding under `WithValue()` — on **every** evaluation, then hands the event to `trace.SpanFromContext(ctx)`. When that span is not recording (in particular, when there is no span in the context at all), `AddEvent` discards the event by the OpenTelemetry contract, so the construction is pure overhead.

This PR returns early in that case:

```go
span := trace.SpanFromContext(ctx)
if !span.IsRecording() {
    return data, nil
}
```

| non-recording `AfterEvaluation` | before | after |
|---|---|---|
| | ~154 ns, 616 B, 4 allocs | **9.6 ns, 0 B, 0 allocs** |

`BenchmarkAfterEvaluationNonRecording` (added) pins the zero-alloc path.

## Why it matters

Evaluations with no span in the context hit this path on every variation call — the common case for services that evaluate flags outside request traces (background work, startup config, non-`Ctx` variation methods without `WithSpans`). A high-volume LaunchDarkly-internal service measured **~3 GB/min of allocation** attributable to this path before working around it locally; this change makes the fix general instead.

## Behavior

- **Recording span present:** byte-identical — same attributes, same event.
- **`WithSpans` variation span:** still ended before the early return (its `End()` precedes the check); existing `TestSpanCreationWithoutParent` covers the no-parent case.
- **Non-recording span:** the only observable difference is that an event the OTel SDK was already discarding is no longer constructed first.

## Verification

`go test ./...` green in `ldotel` (full existing suite), plus the new benchmark.

---

> [!NOTE]
> **Low Risk**
> Observability-only optimization with identical output when spans record; the only change is skipping events that OTel already dropped.
> 
> **Overview**
> **`TracingHook.AfterEvaluation`** now checks `span.IsRecording()` immediately after ending any `WithSpans` variation span and **returns before** building the `feature_flag` event attributes (including JSON flag values when `WithValue()` is enabled). OpenTelemetry would discard those events on non-recording or missing context spans anyway, so this removes per-evaluation allocation on the common path (evaluations outside an active trace).
> 
> Coverage adds a **non-recording span** unit test that asserts `AddEvent` is never called, plus **`BenchmarkAfterEvaluationNonRecording`** to lock in the zero-allocation fast path. Recording-span behavior is unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 04887451c43dceb922589848a2fb20fea4c40b36. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


aaron-zeisler:
Great change! 👍 I'm going to push one more test: the benchmark test is great, but it doesn't test the new AddEvent-skipping functionality.

aaron-zeisler (approved):

Issue #71, opened 2022-05-11
Built-in Attributes break if in 5.x if you have previously been setting them as custom attributes in 4.x


bschaeffer:
**Describe the bug**
We recently became aware of the EOL policy for the go sdk. Since we're running a version of the SDK that has been unsupported for whatever reason for over 1y4m, we decided it was time to upgrade from 4.13 (?) to 5.4. When we did, all targeting by `email` was broken. Under the **4.13** SDK, this behavior worked fine. We set `email` as a custom attribute and our users were able to create flag rules and target variations using that value. When we upgraded to **5.4**, this functionality broke. Variations targeting specific email addresses were ignored. Obviously the DSL has changed, but semantically, we are calling the same APIs.

4.13:

```go
import ldclient "gopkg.in/launchdarkly/go-server-sdk.v4"

atts := map[string]interface{}{"email": "foo@bar.com"}
user := ldclient.NewUser("somekey")
user.Custom = &atts
```

5.4

```go
atts := map[string]interface{}{"email": "foo@bar.com"}
valueMap := ldvalue.CopyArbitraryValueMap(atts)
user := lduser.NewUserBuilder(target.GetKey()).CustomAll(valueMap)
```

There is basically no reason to expect the 4.13 code to work differently from the 5.4 code, but they do. We have identified the solution, which is to call `builder.Email("foo@bar.com")`, but that this call is required or that the other will not work is not apparent at all.

**To reproduce**
* Use the 5.4 SDK
* Set rules for a flag variation targeting a specific email that is different than the default variation
* Set that email as a custom attribute: `builder.Custom("email", ldvalue.String("foo@bar.com"))`
* Evaluate the flag.
* Observe that the variation is not matched and it falls through to the default for the flag.

**Expected behavior**

The old behvior continues to work and the flag is matched on email if its sent as a custom attribute.

**SDK version**
5.4 

**Language version, developer tools**
go 1.18

**OS/platform**
mac/debian

eli-darkly:
I think there may have been something going on with your older code that was not quite what you thought. That is: I don't see how it could have ever worked as you describe.

[...]

PR #429, opened 2026-08-07
feat: conform FDv1 streaming and polling to RETRY spec

tanderson-ld:
## Summary

Brings the FDv1 streaming and polling data sources into conformance with the [RETRY specification](https://github.com/launchdarkly/sdk-specs). Previously-terminal HTTP responses (401, 403, other 4xx that aren't 400/408/429) and TLS/certificate validation failures no longer cause a permanent stop — the data sources engage an extended-regime backoff and continue retrying indefinitely.

- **Streaming:** unexpected failures activate an extended retry curve on the underlying `eventsource` stream (base 5 min → cap 1 h, doubling with jitter). Sixty seconds of continuous healthy operation — measured from the last received event, not from connection establishment — reverts to the normal-regime curve.
- **Polling:** new `pollingStrategy` state machine encoding RETRY §1.4 with a wait floor at `PollInterval` per §1.4.4. Unexpected failures engage the extended regime with `initialDelay = max(configured, PollInterval)`; two consecutive successful polls reset back to the normal regime per §1.8.

FDv2 adoption of RETRY is out of scope for this ticket (deferred to a future epic).

## Public API

Unchanged, and no test-only surface is added.

An earlier revision of this PR exposed extended-regime timing knobs through `<Builder>.Internal()` escape hatches plus a new `testhelpers/datasourcetest` package, so the harness could compress the 5-minute extended delay into an observable window. Reviewers pushed back on forcing test-only knobs into the public API surface just to be testable, and that approach was reverted. The contract tests exercise real production timing instead.

## Behavior change to call out in release notes

**A bad or revoked SDK key no longer fails fast.** Previously a 401 was terminal: the data source stopped, state went to `OFF`, and `MakeCustomClient` returned `ErrInitializationFailed`. Now it returns `ErrInitializationTimeout` once the full `waitFor` elapses, state stays `INITIALIZING` (or `INTERRUPTED` if the client had previously reached `VALID`), and the SDK keeps retrying on the extended curve. Callers matching on `ErrInitializationFailed` to detect a bad key will see different behavior and will block for the whole init timeout rather than returning immediately.

`DataSourceStateOff` is now reserved for explicit shutdown and unparseable configuration.

## Testservice / contract tests

- Declares the `retry-conformance-fdv1-streaming` and `retry-conformance-fdv1-polling` capabilities so the sdk-test-harness can drive the conformance suite.
- No servicedef additions: the compression knobs from the earlier revision were removed along with the public API escape hatches.
- Harness-side tests are launchdarkly/sdk-test-harness#404, now merged to `v2` and shipped in 2.41.0. They are marked long-running and run only under `-enable-long-running-tests`, which this repo's CI does not currently pass, so the retry-conformance suite is exercised out of band for now. Note that declaring the capabilities also deactivates the legacy "do not retry" tests, which are gated on the capability's absence.

## Test plan

- [x] Unit tests in `internal/datasource` pass (streaming + polling + `pollingStrategy`)
- [x] End-to-end tests in `ldclient_end_to_end_test.go` pass
- [x] eventsource dependency is released: v1.14.0 is published, `go.mod` requires it, and CI builds against the released version with no local `replace`
- [x] Full contract suite with `-enable-long-running-tests` against harness `v2` @ 2.41.0: **4931 total, 4922 ran, 9 skipped, 0 failed** (~51 min). All seven retry-conformance subtests ran; the 9 skips are unrelated capabilities (`tls:verify-peer`, `user-type`, `wrapper`), persistence, and the two legacy "do not retry" tests correctly self-skipping.
- [x] End-to-end 401 validation against a fault-injecting origin, at real production timings:
  - **streaming, entry and reset:** 401 → extended regime engaged, first retry at 3m37s (vs the 30 s normal ceiling). After 75 s of healthy operation and a benign close, reconnect in **710 ms** on the normal curve. A fresh 401 re-engages the extended regime from its base rather than from a doubled state.
  - **streaming, ceiling:** sustained 401s climb 3m01s → 8m12s → 12m37s → 31m25s → 55m49s → **49m44s** — doubling off the 5 min base, then clamping at the 1 h ceiling. The sixth rung is the decisive one: an uncapped ladder would have required 80–160 min there. Recovery from the ceiling reconnects in **554 ms** on the normal curve.
  - **polling:** 401 → extended (2m52s, vs the 30 s cadence); still extended after a single success (3m38s); the normal 30 s cadence returns only after the second consecutive success, per §1.8.1.
  - `DataSourceStateOff` was never observed in any run.
- [x] Ladder arithmetic also pinned deterministically against released eventsource v1.14.0 with these exact constants: exact ladder `5m, 10m, 20m, 40m, 1h, 1h, 1h, 1h`; every delay within `[T/2, T]` over 200 jittered trials; reset from the ceiling returns exactly the 1 s normal base and re-arms the extended profile at 5 min; 59 s of healthy operation does **not** reset.

### Known follow-ups (not addressed here)

- Per-SDK CI scheduling for the long-running suite.
- Polling recovery is count-based, and the first success does not shorten the wait. After an outage long enough to reach the 1 h ceiling, returning to the 30 s cadence can therefore take roughly two hours. Measured at the 5 min rung; extrapolated at the ceiling. Streaming is unaffected because its reset is time-based.

---

> [!NOTE]
> **Overview**
> FDv1 streaming and polling now **retry indefinitely** instead of treating 401/403/other 4xx (except 400/408/429) and TLS/cert errors as terminal. Those failures are classified as **unexpected** and use a slower backoff (default 5m, cap 1h, jitter); other errors stay on the normal curve. Status stays `Interrupted` (or `Initializing` if never `Valid`); `Off` is reserved for explicit shutdown or unparseable config. Init on a bad key now times out (`ErrInitializationTimeout`) rather than `ErrInitializationFailed`.
> 
> Streaming uses eventsource **retry profiles** (`eventsource` 1.14) and cancels via request context. Polling adds a `pollingStrategy` state machine with a `PollInterval` wait floor and reset after two consecutive successes.
> 
> Docs for `DataSourceStateOff` and client init errors are updated to match. Testservice advertises `retry-conformance-fdv1-streaming` / `retry-conformance-fdv1-polling`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 576f4d3aa11d459f6f8b4f5ff9382e828c922e8a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/datasource/streaming_data_source.go
```diff
@@ -298,56 +303,80 @@ func (sp *StreamProcessor) subscribe(closeWhenReady chan<- struct{}) {
...
+
+		// Per RETRY §1.2.1: no failure is permanently terminal. Unexpected failures
+		// engage the extended-regime curve; the library keeps retrying at extended
+		// cadence until a healthy-op reset (retryResetInterval of continuous
+		// connection) reverts.
+		result := es.StreamErrorHandlerResult{CloseNow: false}
+		if class == FailureClassUnexpected {
+			result.ActivateCurve = extendedCurve
+		}
+		return result
```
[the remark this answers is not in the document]
tanderson-ld: Fixed in 655f9c6c (SDK side) + launchdarkly/eventsource#71 (SSE library side).

The SDK now builds the streaming request via `http.NewRequestWithContext` with a context that `Close()` cancels. Under the current eventsource (v1.13.0), this interrupts any in-flight `Do()` immediately -- net/http honors `req.Context()`. Full retry-sleep interruption arrives once eventsource #71 releases and this PR's `go.mod` is bumped: PR #71 adds context observation in `SubscribeWithRequestAndOptions`'s retry-sleep select and installs a `context.AfterFunc` bridge from the request context to `Stream.Close`, so cancellation covers the whole Subscribe lifetime rather than just pre-connect.

The pre-existing gap was actually latent well before this PR (see PR #71's description) -- SDK-2788 widened the failure classes that hit the uninterruptible retry loop (401/403 previously self-terminated via the errorHandler's `CloseNow: true` return), which is what made this customer-visible now.

internal/datasource/helpers.go
```diff
@@ -20,53 +23,114 @@ func (e httpStatusError) Error() string {
...
-) bool {
-	if statusCode > 0 && !isHTTPErrorRecoverable(statusCode) {
-		loggers.Errorf("Error %s (giving up permanently): %s", errorContext, errorDesc)
-		return false
+	willRetryMessage string,
+) FailureClass {
+	class := classifyHTTPFailure(statusCode)
+	if class == FailureClassUnexpected {
+		loggers.Errorf("Error %s (%s; will continue retrying with extended backoff): %s",
+			errorContext, willRetryMessage, errorDesc)
```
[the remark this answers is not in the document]
tanderson-ld: Fixed in 655f9c6c. `classifyAndLogHTTPFailure` and `classifyAndLogTransportFailure` now both emit `Warnf` and drop the hardcoded "; will continue retrying with extended backoff" suffix. Rationale:

- Under RETRY §1.2.1 no failure is permanent, so every per-attempt log describes a temporary condition -- per CLM 1.1.4 that's Warn level, not Error.
- The suffix contradicted the caller-provided `willRetryMessage` (polling's "will retry at next scheduled poll interval") and doesn't survive the level change.

Regime state is instead surfaced by a one-time Info log at the transition into extended for both streaming and polling: `Classified failure as UNEXPECTED; engaging extended backoff.` -- mirrors the sister Java SDK's approach.

The resulting Warn-level log format for both classifications is byte-for-byte identical to the pre-RETRY "recoverable" log branch (e.g. `Error on polling request (will retry at next scheduled poll interval): HTTP error 401`). No new log shapes, no new Error-level entries; the transition Info line is the only additive.

internal/datasource/polling_strategy.go
```diff
@@ -0,0 +1,101 @@
...
+// data source. It owns:
+//
+//   - the attempts counter n (RETRY §1.4);
+//   - the current regime's (initialDelay, maxDelay), toggled by classification
+//     (RETRY §1.5–§1.7 via the caller's FailureClass);
+//   - the two-consecutive-success reset gate (RETRY §1.8 polling binding);
+//   - jitter (RETRY §1.4.3);
+//   - the PollInterval wait floor (RETRY §1.4.4 polling override).
+//
+// All state is owned and mutated by the polling run() goroutine only — no
```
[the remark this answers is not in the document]
tanderson-ld: Swept across all PR-touched files in 655f9c6c: em/en dashes to `--`, minus signs to `-`, arrows to ASCII forms.

internal/datasource/polling_strategy.go
```diff
@@ -0,0 +1,109 @@
...
+		s.n = 1
+		s.initialDelay = s.extendedInitialPollInterval
+		if s.initialDelay < s.normalInterval {
+			s.initialDelay = s.normalInterval
+		}
+		s.maxDelay = extendedPollMaxDelay
+		if s.maxDelay < s.normalInterval {
+			s.maxDelay = s.normalInterval
+		}
+		return
```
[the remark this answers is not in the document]
tanderson-ld: Confirmed real bug, fixed in 655f9c6c.

Prior detection relied on `s.initialDelay == s.normalInterval`. That equality holds AFTER the transition path clamps `initialDelay` up to `max(extendedInitialPollInterval, PollInterval)` whenever `PollInterval >= extendedInitialPollInterval` -- including the default 5min/5min combo. So every subsequent unexpected failure re-fired the transition branch, reset `n` to 1, and defeated RETRY §1.4.1's `T = initialDelay * 2^(n-1)` doubling.

Replaced with an explicit `inExtended bool` on `pollingStrategy`. `OnFailure` now returns `true` iff this call transitioned from normal into extended, so the caller can emit the one-time Info log added for Thread 2. `OnSuccess`'s two-consecutive-success reset also clears `inExtended` so post-reset unexpected failures re-transition properly.

Regression tests added:

- `ExtendedDoublingWhenClampedToPollInterval` -- 5min/5min config, five unexpected failures, asserts `n` advances 1->5 and NextWait grows accordingly (would have stayed at n=1 with the old detection).
- `OnFailureReturnsTrueOnceOnTransition` and `...WhenClamped` -- returns true only on the first unexpected, false on subsequent.
- `OnSuccessResetAllowsRetransition` -- post-reset unexpected returns true again.
- `ExtendedDoublingWithModeratePollInterval` -- 10min/5min config, doubling engages from n=2 and reaches ceiling at n=4.
- `ExtendedRegimeCollapsesWhenPollIntervalExceedsExtendedCeiling` -- 2h/5min config, both `initialDelay` and `maxDelay` clamp up to PollInterval, every wait is exactly 2h. Locks in the invariant that extended regime is behaviorally identical to normal cadence in this config.

The existing `ExtendedInitialClampedToPollInterval` test didn't catch the bug because it drove only one failure and its 1h/5min config happened to produce identical NextWait output with or without doubling (maxDelay == PollInterval == 1h). Analysis in the discussion thread.

internal/datasource/polling_strategy_test.go
```diff
@@ -0,0 +1,204 @@
...
+	assert.Equal(t, time.Hour, s.maxDelay)
+	// All subsequent waits collapse to PollInterval.
+	assert.Equal(t, time.Hour, s.NextWait())
+}
+
+// After prior normal-classified failures, the first unexpected failure MUST
+// reset n to 1 so the first extended-regime wait uses the new initialDelay
+// (5min) directly, not initialDelay * 2^k where k is the count of prior
+// normal failures. Guards against conflating the two roles of a single
+// counter — formula input (this field, n; resets on regime transition per
```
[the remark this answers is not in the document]
tanderson-ld: Swept in 655f9c6c along with Thread #3.

internal/datasource/streaming_data_source.go
```diff
@@ -298,56 +306,91 @@ func (sp *StreamProcessor) subscribe(closeWhenReady chan<- struct{}) {
...
-		return es.StreamErrorHandlerResult{CloseNow: false}
+
+		// Per RETRY §1.2.1: no failure is permanently terminal. Unexpected failures
+		// engage the extended-regime profile; the library keeps retrying at extended
+		// cadence until a healthy-op reset (retryResetInterval of continuous
+		// connection) reverts.
+		result := es.StreamErrorHandlerResult{CloseNow: false}
+		if class == FailureClassUnexpected {
+			if !loggedActivatedExtended {
+				sp.loggers.Info("Classified failure as UNEXPECTED; engaging extended backoff.")
```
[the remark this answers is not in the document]
tanderson-ld: Not worth narrowing.

internal/datasource/streaming_data_source.go
```diff
@@ -259,15 +271,17 @@ func (sp *StreamProcessor) consumeStream(stream *es.Stream, closeWhenReady chan<
 				sp.setInitializedAndNotifyClient(true, closeWhenReady)
 			}
 
-		case <-sp.halt:
-			stream.Close()
+		case <-sp.streamReqCtx.Done():
```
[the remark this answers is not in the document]
tanderson-ld: I think this is fixed by https://github.com/launchdarkly/eventsource/pull/71, which still does need to merge and be used in this PR.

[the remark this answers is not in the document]
tanderson-ld: eventsource v1.14.0 (which ships #71) is now consumed in c355d5eb. `StreamProcessor.Close` already cancels its `streamReqCtx`, and v1.14.0 teaches `Subscribe` to observe that context as a stream-lifetime cancel signal, so an in-flight retry timer is now interrupted instead of blocking teardown until it fires. Resolving.

internal/datasource/helpers.go
```diff
@@ -20,53 +23,100 @@ func (e httpStatusError) Error() string {
 	return e.Message
 }
 
-// Tests whether an HTTP error status represents a condition that might resolve on its own if we retry,
-// or at least should not make us permanently stop sending requests.
-func isHTTPErrorRecoverable(statusCode int) bool {
+// FailureClass categorizes a data source failure per RETRY §1.5--§1.7. Under the
```
beekld: here and elsewhere, we shouldn't put spec numbers in comments or code

internal/datasource/helpers.go
```diff
@@ -20,53 +23,100 @@ func (e httpStatusError) Error() string {
...
-) bool {
-	if statusCode > 0 && !isHTTPErrorRecoverable(statusCode) {
-		loggers.Errorf("Error %s (giving up permanently): %s", errorContext, errorDesc)
-		return false
-	}
-	loggers.Warnf("Error %s (%s): %s", errorContext, recoverableMessage, errorDesc)
-	return true
+	willRetryMessage string,
+) FailureClass {
+	loggers.Warnf("Error %s (%s): %s", errorContext, willRetryMessage, errorDesc)
```
beekld: optional: Would it make sense to still use `Errorf` for the `FailureClassUnexpected` case?

internal/datasource/helpers_test.go
```diff
@@ -12,19 +16,47 @@ func TestHTTPStatusError(t *testing.T) {
 	assert.Equal(t, "message", error.Error())
 }
 
-func TestIsHTTPErrorRecoverable(t *testing.T) {
+// classifyHTTPFailure per RETRY §1.6: 400, 408, 429 are normal (transient
+// server-side conditions); all other 4xx are unexpected (durable client-side
+// misconfiguration); 5xx and everything else are normal.
+func TestClassifyHTTPFailure(t *testing.T) {
 	for i := 400; i < 500; i++ {
```
beekld, line 23: lol, this is excessive, but existing code, so i wouldn't ask you to change it.

but i do wonder about the choice not to test 3xx, 1xx, etc.

internal/datasource/polling_data_source.go
```diff
@@ -80,10 +92,13 @@ func newPollingProcessor(
 func (pp *PollingProcessor) Start(closeWhenReady chan<- struct{}) {
 	pp.loggers.Infof("Starting LaunchDarkly polling with interval: %+v", pp.pollInterval)
 
-	ticker := newTickerWithInitialTick(pp.pollInterval)
+	// Fires immediately for the first poll; Reset after each iteration to schedule
+	// the next. Under RETRY (SDK-2775), the interval between polls is dynamic per
```
beekld: we shouldn't put a JIRA ticket number in the comment. same elsewhere in this PR

internal/datasource/polling_data_source_test.go
```diff
@@ -141,10 +141,15 @@ func testPollingProcessorRecoverableError(t *testing.T, err error, verifyError f
 	})
 }
 
-func TestPollingProcessorUnrecoverableErrors(t *testing.T) {
+// Under the RETRY spec (SDK-2775), previously-terminal 4xx errors (401, 403, 404,
```
beekld: This comment should be rewritten to focus on what the code does, rather than what it used to do.

(and avoid the ticket name)

internal/datasource/polling_strategy.go
```diff
@@ -0,0 +1,115 @@
...
+		float64(s.initialDelay)*math.Pow(2, float64(s.n-1)),
+		float64(s.maxDelay),
+	))
+	var jitter time.Duration
+	if halfT := int64(t / 2); halfT > 0 {
+		jitter = time.Duration(s.rng.Int63n(halfT))
+	}
+	wait := t - jitter
+	if wait < s.normalInterval {
+		wait = s.normalInterval
```
beekld, line 111: i worry that, depending on how this is configured, if it hits this min often, won't it undo the effect of the jitter? i'm not sure what to do in that case though

internal/datasource/polling_strategy_test.go
```diff
@@ -0,0 +1,317 @@
...
+
+	assert.Equal(t, time.Hour, s.initialDelay)
+	assert.Equal(t, time.Hour, s.maxDelay)
+	// All subsequent waits collapse to PollInterval.
+	assert.Equal(t, time.Hour, s.NextWait())
+}
+
+// After prior normal-classified failures, the first unexpected failure MUST
+// reset n to 1 so the first extended-regime wait uses the new initialDelay
+// (5min) directly, not initialDelay * 2^k where k is the count of prior
```
beekld: this comment is pretty hard to follow

internal/datasource/streaming_data_source_test.go
```diff
@@ -248,10 +249,15 @@ func TestStreamProcessorRecoverableErrorsCauseStreamRestart(t *testing.T) {
 	})
 }
 
-func TestStreamProcessorUnrecoverableErrorsCauseStreamShutdown(t *testing.T) {
+// Under the RETRY spec (SDK-2775), 401 / 403 / other 4xx are no longer terminal --
+// they engage an extended-regime backoff but keep retrying indefinitely. The SDK
+// transitions to Interrupted (not Off) and does not close the initialization
+// channel. This test replaces the pre-RETRY TestStreamProcessorUnrecoverableErrors
```
beekld: This comments seems excessive, and we don't need to document what the old test was.

beekld (changes requested):
the code seems fine to me. the tests have good coverage. but i think we need a pass on the comments to bring them in line with our standards

internal/datasource/helpers.go
tanderson-ld: Done in 28d906e9 — swept `RETRY §X.Y` citations and ticket references from all touched comments across the diff.

internal/datasource/helpers.go
tanderson-ld: Already implemented in 5d760369 — `FailureClassUnexpected` logs at Error level while normal failures log at Warn. Behavior matches your suggestion.

internal/datasource/helpers_test.go
tanderson-ld: Intent was to focus on the 4xx boundary since that's where the classification split (normal vs unexpected) lives. 1xx/3xx don't reach this path in practice — the caller only invokes `classifyHTTPFailure` for non-2xx responses, and HTTP 1xx would be handled by the client transport before we see it. Leaving as-is per your note.

internal/datasource/polling_data_source.go
tanderson-ld: Done in 28d906e9 — removed the ticket reference here and the other citation/ticket instances across the diff.

internal/datasource/polling_data_source_test.go
tanderson-ld: Rewrote in 28d906e9 to describe current behavior only; dropped the "Replaces X, which asserted the old permanent-stop behavior" framing. Same treatment applied to the streaming and end-to-end test equivalents.

internal/datasource/polling_strategy.go
tanderson-ld: Traced through — the floor binds and squashes jitter in specific bounded cases:

- **Normal regime** (`T ≡ PollInterval`): jitter always squashes by design — the customer configured that interval and shouldn't get shorter polls.
- **Extended regime, `PollInterval < extendedBase`** (defaults: 30s / 5min, i.e. most customers): jitter operates fully at every attempt.
- **Extended regime, `extendedBase < PollInterval < extendedPollMaxDelay`**: jitter squashes at n=1 (fully) and at n=2 only at the exact lower edge of the range. From n=3 onward `T > 2×PollInterval` and the floor never binds.
- **Extended regime, `PollInterval > extendedPollMaxDelay` (1h)**: extended and normal regimes collapse to the same behavior — no jitter at all, but by construction.

The framing that eased my worry: in the cases where the floor binds, behavior degenerates to *exactly the pre-PR polling behavior* — a fixed `PollInterval` cadence with no jitter. So this PR strictly improves herd-avoidance for the common configuration and doesn't regress for the edge cases. The alternatives I considered (jitter-add, scaled jitter, n=1-only floor) either double the effective backoff or violate the flat `PollInterval` floor.

Case B and Case C bounds are exercised by `TestPollingStrategy_ExtendedDoublingWithModeratePollInterval` and `TestPollingStrategy_ExtendedRegimeCollapsesWhenPollIntervalExceedsExtendedCeiling`. Leaving as-is — happy to revisit if you spot a formulation I missed.

internal/datasource/polling_strategy_test.go
tanderson-ld: Rewrote in 28d906e9 — dropped the abstract "two roles of a single counter" framing and the mid-sentence `2^k` construction. Now leads with the invariant and gives one concrete example (5min → 20min inflation from a "normal, normal, unexpected" sequence).

internal/datasource/streaming_data_source_test.go
tanderson-ld: Trimmed in 28d906e9 to three lines describing current behavior; dropped the "replaces the pre-RETRY test" historical framing. Same treatment applied to the polling and end-to-end equivalents.

internal/datasource/polling_strategy.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// OnFailure updates the strategy state after a failed poll. An Unexpected
+// classification engages the extended regime for subsequent waits;
+// initialDelay is clamped to at least PollInterval so the extended regime
+// is never faster than the customer-configured normal cadence.
+//
+// On the transition from normal into extended regime, n is reset to 1 so
+// that the first extended-regime wait uses the new initialDelay. Returns
+// true iff this call transitioned the strategy from normal into extended
```
jsonbailey:
```suggestion
// true if this call transitioned the strategy from normal into extended
```

internal/datasource/streaming_data_source.go
```diff
@@ -36,22 +37,27 @@ import (
...
-	deleteEvent              = "delete"
-	streamReadTimeout        = 5 * time.Minute // the LaunchDarkly stream should send a heartbeat comment every 3 minutes
-	streamMaxRetryDelay      = 30 * time.Second
-	streamRetryResetInterval = 60 * time.Second
-	streamJitterRatio        = 0.5
-	defaultStreamRetryDelay  = 1 * time.Second
+	putEvent    = "put"
+	patchEvent  = "patch"
+	deleteEvent = "delete"
+	// The LaunchDarkly stream should send a heartbeat comment every 3 minutes.
```
jsonbailey, line 53: This comment was here previously but should we adjust this to say 5 minutes or set the timeout to be 3 minutes?

jsonbailey (approved):
Minor nits that I won't block but should be easily fixed

internal/datasource/streaming_data_source.go
tanderson-ld: I think this is "the FD server _SHOULD_ send a heartbeat every 3 minutes, so 5 minutes as the timeout is fine.

beekld (approved):

PR #439, opened 2026-09-01
ci: use shared sdk-go-versions workflow

keelerm84:
Replaces this repo's copy of the Go version check with a call to the shared `sdk-go-versions` reusable workflow, deleting the inline shell entirely.

**Why this supersedes the previous approach**

The earlier revision of this PR hardened the inline shell in place (the port of launchdarkly/ld-relay#841, SEC-9481) — quoting the untrusted endoflife.date payload into an env var, adding `jq` extraction and a version regex. That fixed the injection; it did not remove the surface. Calling the shared workflow removes the shell from this repo altogether, so the script-injection class is structurally unavailable here rather than merely patched. 78 lines of inline workflow (91 after the earlier hardening revision) become 21 lines, none of which is a `run:` step.

The shared workflow pins the composite action to the released `go-eol-versions-v0.1.0`; consumers track the workflow itself at `@main`, matching how this org already consumes `sdk-stale` and `dependency-scan` (release-please does not version reusable workflows in gh-actions — see launchdarkly/gh-actions#51).

**Inputs**

- `branches: '["v7"]'` — the only maintained release branch for this repo, matching the previous `strategy.matrix.branch` value.

Everything else is left at its default: `precision` (`cycle`), `versions_file` (`.github/variables/go-versions.env`), `add_paths`, and `pr_title_prefix` already match what this repo needs.

**Behaviour**

Intended to be unchanged: same `0 17 * * *` schedule plus `workflow_dispatch`, same `.github/variables/go-versions.env` updated, same bot author/committer and PR shape. Two fixes the shared version carries that the inline copy did not:

- A penultimate-only upstream release now opens a PR. The inline gate compared only `latest`, so a new penultimate with an unchanged latest was silently dropped.
- The bot branch name includes both versions, so a latest-only and a penultimate-only bump can no longer collide on one branch.

**Verification**

- YAML parses and resolves to a single reusable-workflow call: `python3 -c "import yaml; d=yaml.safe_load(open('.github/workflows/check-go-versions.yml')); print(d['jobs'])"` → one `check-go-versions` job with `uses:`, `with: {branches: '["v7"]'}`, and `contents: write` / `pull-requests: write`.
- No inline shell remains: `grep -c 'run:'` → 0.
- No remnants of the old implementation: `grep -c 'fromJSON\|JamesIves\|peter-evans\|sed -i'` → 0.
- `git diff --stat` against `v7` touches only `.github/workflows/check-go-versions.yml` (11 insertions, 68 deletions).
- The workflow was not dispatched from this repo; the shared workflow is verified end to end in launchdarkly/go-test-helpers ([run 33534426032](https://github.com/launchdarkly/go-test-helpers/actions/runs/33534426032)).

Tracked by SDK-3023.
Shared workflow: launchdarkly/gh-actions#113.

**Requirements**

- [x] I have added test coverage for new or changed functionality — n/a — workflow-only change

---

> [!NOTE]
> **Overview**
> **Replaces the in-repo Go version automation** with a call to `launchdarkly/gh-actions` reusable workflow `sdk-go-versions.yml@main`, shrinking `check-go-versions.yml` from two jobs and ~78 lines of inline steps to a single job with no local `run:` blocks.
> 
> The removed implementation fetched supported versions from endoflife.date, updated `.github/variables/go-versions.env` via `sed`, and opened PRs with `peter-evans/create-pull-request` on branch `v7`. **Behaviour is intended to stay the same** (daily `0 17 * * *` schedule, `workflow_dispatch`, same env file and bot PRs) via `with: branches: '["v7"]'` and shared defaults for paths and titles.
> 
> **Why:** Centralizing logic in gh-actions removes shell/script-injection surface from this repo and picks up shared fixes (e.g. PRs when only penultimate changes, distinct bot branch names per bump).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ac80e11296c2e93e0590e7ab5cfb722313bc9e38. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/check-go-versions.yml
+++ b/.github/workflows/check-go-versions.yml
@@ -7,85 +7,15 @@ on:
 permissions: {}
 
 jobs:
-  check-go-eol:
-    permissions:
-      contents: read
-    runs-on: ubuntu-latest
-    outputs:
-      latest: ${{ steps.parse.outputs.latest }}
-      penultimate: ${{ steps.parse.outputs.penultimate }}
-    timeout-minutes: 2
-    steps:
-        # Perform a GET request to endoflife.date for the Go language. The response
-        # contains all Go releases; we're interested in the 0'th and 1'th (latest and penultimate.)
-      - name: Fetch officially supported Go versions
-        uses: JamesIves/fetch-api-data-action@396ebea7d13904824f85b892b1616985f847301c # v2.2.4
-        with:
-          endpoint: https://endoflife.date/api/go.json
-          configuration: '{ "method": "GET" }'
-          debug: true
-        # Parse the response JSON and insert into environment variables for the next step.
-      - name: Parse officially supported Go versions
-        id: parse
-        env:
-          FETCH_API_DATA: ${{ env.fetch-api-data }}
-        run: |
-          set -euo pipefail
-          latest=$(jq -r '.[0].cycle' <<< "$FETCH_API_DATA")
-          penultimate=$(jq -r '.[1].cycle' <<< "$FETCH_API_DATA")
-          for version in "$latest" "$penultimate"; do
-            if [[ ! "$version" =~ ^[0-9]+\.[0-9]+(\.[0-9]+)?$ ]]; then
-              echo "Unexpected Go version string from endoflife.date: $version" >&2
-              exit 1
-            fi
-          done
-          echo "latest=$latest" >> "$GITHUB_OUTPUT"
-          echo "penultimate=$penultimate" >> "$GITHUB_OUTPUT"
-
-
-  create-prs:
+  check-go-versions:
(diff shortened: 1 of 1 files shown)


keelerm84:
Tracked by [SDK-3023](https://launchdarkly.atlassian.net/browse/SDK-3023), which carries the full context: the other repos in this sweep, the shared-action follow-up, and the recommended merge order.

[SDK-3023]: https://launchdarkly.atlassian.net/browse/SDK-3023?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

jsonbailey (approved):

PR #440, opened 2026-09-01
feat: deprecate FDv2 payload filter builder methods

beekld:
Deprecates the FDv2-specific `PayloadFilter` builder methods, since payload filtering is not supported with the FDv2 data system.

Closes [SDK-2972](https://launchdarkly.atlassian.net/browse/SDK-2972), part of [SDK-2575](https://launchdarkly.atlassian.net/browse/SDK-2575).

- `PollingDataSourceBuilderV2.PayloadFilter` and `StreamingDataSourceBuilderV2.PayloadFilter` now carry a `Deprecated:` note pointing at the FDv1 equivalents.
- No behavior change: the methods still work, and the FDv1 builders (including `FDv1PollingDataSourceBuilderV2`, used for FDv1 fallback) are untouched.
- Removal (and dropping the `dataSystem.payloadFilter` wiring in the test service) is left for a follow-up once the contract test harness no longer sends that field.

<details>
<summary>Implementation details</summary>

The contract test harness removed the FDv2 payload filter scenarios (launchdarkly/sdk-test-harness#431), so the public FDv2 filter surface is going away across SDKs. This PR is the deprecation step only, keeping the change non-breaking; the test service still maps `DataSystem.PayloadFilter` into these builders, which is why they aren't removed yet.

Doc-comment-only change, so no new tests. Verified with `go build ./...`, `go vet ./...`, `make lint` (0 issues) and `go test ./ldcomponents/...`.

</details>

Link to Devin session: https://app.devin.ai/sessions/ed604c6db94d452ea9f2dc35d6acc398
Open in Devin Desktop: https://app.devin.ai/desktop/session/ed604c6db94d452ea9f2dc35d6acc398?variant=devin
Requested by: @beekld

[SDK-2972]: https://launchdarkly.atlassian.net/browse/SDK-2972?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SDK-2575]: https://launchdarkly.atlassian.net/browse/SDK-2575?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Overview**
> **FDv2** polling and streaming data source builders now treat `PayloadFilter` as deprecated: godoc notes that payload filtering is unsupported on the FDv2 data system and the methods will be removed later.
> 
> On `Build`, configuring any payload filter (including an empty string) no longer returns an error for `""`; instead the SDK logs a **warn** that filtering is not supported and will stop being applied in a future release. Filter values are still passed through to the underlying config for now.
> 
> Tests for both builders were updated to expect success with an empty filter and to assert the deprecation warning is emitted only when a filter is set.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3075cef7d71c2370b9da60a885d2016d7592375a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ldcomponents/polling_data_source_builder_v2.go
+++ b/ldcomponents/polling_data_source_builder_v2.go
@@ -66,8 +66,7 @@ func (b *PollingDataSourceBuilderV2) forcePollInterval(
 // Evaluations for flags that aren't part of the filtered environment will return default values.
 //
 // Deprecated: Payload filtering is not supported with the FDv2 data system and this method will be
-// removed in a future release. There is no replacement: payload filtering is only available with
-// the FDv1 data source.
+// removed in a future release.
 func (b *PollingDataSourceBuilderV2) PayloadFilter(filterKey string) *PollingDataSourceBuilderV2 {
 	b.filterKey = ldvalue.NewOptionalString(filterKey)
 	return b
--- a/ldcomponents/polling_data_source_builder_v2.go
+++ b/ldcomponents/polling_data_source_builder_v2.go
@@ -10,6 +10,9 @@ import (
 	"github.com/launchdarkly/go-server-sdk/v7/subsystems"
 )
 
+const deprecatedPayloadFilterMessage = "Payload filtering is not supported with the FDv2 data system;" +
+	" the configured payload filter will stop being applied in a future release"
+
 // PollingDataSourceBuilderV2 provides methods for configuring the polling data source.
 type PollingDataSourceBuilderV2 struct {
 	pollInterval time.Duration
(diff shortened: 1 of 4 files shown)


tanderson-ld (approved):

ldcomponents/polling_data_source_builder_v2.go
```diff
@@ -64,6 +64,9 @@ func (b *PollingDataSourceBuilderV2) forcePollInterval(
 // a smaller, known subset - then a filter may be setup in LaunchDarkly, and the filter's key specified here.
 //
 // Evaluations for flags that aren't part of the filtered environment will return default values.
+//
+// Deprecated: Payload filtering is not supported with the FDv2 data system and this method will be
```
keelerm84: I wonder if we shouldn't move this comment towards the top for better visibility, particularly since this function effectively does nothing.

For SDKs that make it easy, we should also log a warning if they are using these methods. Most SDKs won't have a logger in the builder, so we may want to look at logging somewhere post construction.

ldcomponents/polling_data_source_builder_v2.go
```diff
@@ -75,8 +78,11 @@ func (b *PollingDataSourceBuilderV2) PayloadFilter(filterKey string) *PollingDat
 // Build is called internally by the SDK.
 func (b *PollingDataSourceBuilderV2) Build(context subsystems.ClientContext) (subsystems.DataSynchronizer, error) {
 	filterKey, wasSet := b.filterKey.Get()
-	if wasSet && filterKey == "" {
-		return nil, errors.New("payload filter key cannot be an empty string")
+	if wasSet {
+		if filterKey == "" {
+			return nil, errors.New("payload filter key cannot be an empty string")
+		}
```
keelerm84: Since payload filters don't do anything on fdv2, maybe this shouldn't be an error any more? We can just silently ignore the empty string because it was never going to do anything anyway?

keelerm84 (approved):

PR #454, opened 2026-09-17
ci: add nightly contract tests with long-running tests enabled

tanderson-ld:
**Requirements**

- [x] I have followed the repository's [pull request submission guidelines](../blob/v7/CONTRIBUTING.md#submitting-pull-requests)
- [ ] I have added test coverage for new or changed functionality — n/a, CI config only
- [ ] I have validated my changes against all supported platform versions — n/a, single Go version by design

**Related issues**

Tracked in Jira. The issue key is omitted on purpose so that merging this does not auto-close the story before the follow-up fix PR.

The test harness gates a set of RETRY-conformance and FDv2 tests behind `-enable-long-running-tests`, because they wait on real backoff timing and take about an hour. Per-PR CI does not pass that flag, so none of them have ever run for this SDK. This adds a nightly job that does — 11 test groups newly covered. It is the Go counterpart to java-core's nightly.

> [!WARNING]
> **This job is expected to fail on exactly one test** until a follow-up PR in this repo fixes it: `streaming/fdv2/permanent fallback with recovery`. On permanent synchronizer removal the SDK never resets its data source status, so it does not fall back to a healthy synchronizer before the init timeout. Pre-existing — it reproduces on 7.15.6, so not a RETRY regression. Deliberately not suppressed: a nightly that goes red on a real defect is the point. **More than one failure means something else broke** — check the uploaded JUnit results.

Choices worth noting:

- Both harness majors float (`v2`, `v3`), matching `common_ci.yml`, so nightly and per-PR CI resolve the same harness.
- Single Go version (`latest`, from the shared version-matrix workflow). RETRY timing is not Go-version-sensitive, and per-PR CI already runs the fast suite on all three.
- Persistence tests off, so no `persistent-stores` step.
- `timeout-minutes: 120`. java-core's equivalent job measures 59–65 min.
- Two **empty** suppression files, so adding a suppression later is a one-line change. They must exist even when empty — the harness hard-errors on a missing `-skip-from` path.

- **Suppressing the known failure** — rejected; it defeats the purpose. java-core's three suppressions from February were never revisited.
- **Splitting long-running into its own job** — not worth duplicating build and service startup. Revisit if it proves flaky.
- **A `concurrency` group** — unnecessary, since each run gets its own ephemeral VM, and it would queue the dispatch runs used to validate this.
- **Porting java-core's Slack notification** — that webhook has never worked end to end. Fix it there first, then propagate one working solution.

**Additional context**

This cannot be validated before merge, because a new workflow file is not dispatchable until it is on the default branch. After merge:

```
gh workflow run nightly-contract-tests.yml --repo launchdarkly/go-server-sdk
```

Expect zero `use -enable-long-running-tests to run` skips, and exactly one failure.

---

> [!NOTE]
> **Overview**
> Adds a **nightly GitHub Actions workflow** (4:00 UTC cron plus manual dispatch) that runs the SDK contract harness the same way as per-PR CI—build/start the test service, then **v2 and v3** harness runs—but turns on **`-enable-long-running-tests`** so RETRY/backoff and related scenarios that take ~an hour actually execute.
> 
> Compared to **`common_ci.yml` contract tests**, the job uses **latest Go only** (via the shared version matrix), **disables persistence tests** (no persistent-stores setup), passes **`-skip-from`** to empty suppression files under `testservice/`, writes **JUnit** under `junit/`, enables **debug logging**, and sets a **120-minute** job timeout. It always uploads **JUnit artifacts** and test-service logs on completion.
> 
> The workflow is expected to stay red on one known FDv2 failure until a follow-up fix lands; more than one failure would indicate a broader regression.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2aab4271b63c9a7147959e2ba70b24134e89225f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


testservice/test-suppressions-fdv2.txt
tanderson-ld, line 1: Intentionally empty

testservice/test-suppressions.txt
tanderson-ld, line 1: Intentionally empty

.github/workflows/nightly-contract-tests.yml
```diff
@@ -0,0 +1,86 @@
...
+          stop_service: 'false'
+          enable_persistence_tests: 'false'
+          extra_params: '-skip-from=testservice/test-suppressions.txt -enable-long-running-tests -junit junit/contract-tests-v2.xml'
+
+      - name: Run SDK contract test v3
+        uses: launchdarkly/gh-actions/actions/contract-tests@contract-tests-v1
+        with:
+          test_service_port: '10000'
+          version: 'v3'
+          branch: 'v3'
```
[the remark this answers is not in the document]
tanderson-ld: Checked this one — `v3` is a live branch on `sdk-test-harness`, and the `branch` input is a git ref rather than a tag specifically, so the fetch resolves fine:

```
curl .../sdk-test-harness/v3/downloader/run.sh  ->  HTTP 200
git ls-remote --heads origin                    ->  refs/heads/v3
```

Ran the whole path to be sure: it resolved `v3` to `v3.2.0-alpha.10`, downloaded the asset, and executed the binary.

Setting it is deliberate rather than incidental. The v2 and v3 copies of `run.sh` differ, and v3's is the more robust one:

- v2 passes `GITHUB_TOKEN`, but the public download URL drops the `Authorization` header when it redirects to the asset host, so the token never raises the rate limit on the download itself. The v3 script resolves the asset through the releases API instead, which keeps the token in play.
- v3 wraps every request in `--retry 5 --retry-delay 2`; v2 retries only the final download.

The verification run above took that authenticated path (`Downloading https://api.github.com/.../releases/assets/...`), which exists only in the v3 script.

One part of this is a fair catch, though: `common_ci.yml` does leave `branch` at the `v2` default while setting `version: v3`. That works, but it means per-PR CI downloads a v3 harness using the older script and misses both fixes above. Worth a follow-up PR — leaving it out of scope here.

.github/workflows/nightly-contract-tests.yml
```diff
@@ -0,0 +1,86 @@
...
+          token: ${{ secrets.GITHUB_TOKEN }}
+          debug_logging: 'true'
+          enable_persistence_tests: 'false'
+          extra_params: '-skip-from=testservice/test-suppressions-fdv2.txt -enable-long-running-tests -junit junit/contract-tests-v3.xml'
+
+      - name: Upload contract test results
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: contract-test-results
```
joker23: question: where are these test artifacts being updated?

joker23 (approved):

.github/workflows/nightly-contract-tests.yml
tanderson-ld: I believe onto the job itself.  See example here: https://github.com/launchdarkly/go-server-sdk/actions/runs/35638104882

.github/workflows/nightly-contract-tests.yml
```diff
@@ -0,0 +1,86 @@
...
+
+      - name: Run SDK contract test v2
+        uses: launchdarkly/gh-actions/actions/contract-tests@contract-tests-v1
+        with:
+          test_service_port: '10000'
+          token: ${{ secrets.GITHUB_TOKEN }}
+          debug_logging: 'true'
+          stop_service: 'false'
+          enable_persistence_tests: 'false'
+          extra_params: '-skip-from=testservice/test-suppressions.txt -enable-long-running-tests -junit junit/contract-tests-v2.xml'
```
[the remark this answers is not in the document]
tanderson-ld: The flag does exist on v2. From `params.go` at tag `v2.41.0`:

```go
fs.BoolVar(&c.enableLongRunningTests, "enable-long-running-tests", false,
    "enable tests that take a long time to run (10+ seconds)")
```

It has to, because v2 is where these tests live. The gate that skips them is in the v2 framework too:

```go
func (t *T) LongRunning() {
    if !t.env.config.EnableLongRunningTests {
        t.SkipWithReason("use -enable-long-running-tests to run")
    }
}
```

Those `use -enable-long-running-tests to run` skips are exactly what shows up in this SDK's per-PR CI today for the six RETRY groups, and enabling them on v2 is the entire point of this PR. If the flag were unknown to v2, the change would be pointless rather than merely broken.

Verified by running it rather than reasoning about it — harness `v2.41.0` with this step's exact params:

```
Test Summary: 4931 total, 9 skipped, 4922 ran
All tests passed
v2_EXIT=0
```

Zero `use -enable-long-running-tests to run` skips remained, and the v3 step then ran and produced the single expected FDv2 failure — so the v2 step neither aborted nor skipped the v3 suite.

On the java-core comparison: that was true until 2026-09-10, but [java-core#212](https://github.com/launchdarkly/java-core/pull/212) added the flag to its v2 step for this same reason. Its current v2 step reads:

```yaml
extra_params: '-skip-from=lib/sdk/server/contract-tests/test-suppressions.txt -enable-long-running-tests -junit junit/contract-tests-v2.xml'
```

